PR #43, opened 2026-04-23
Fix XSS vulnerabilities

marcotama:
Hi,

I came across `calrs` yesterday, when someone mentioned it in a comment on HN.

I like this and would like to use it, but am a bit too obsessive about security, and am wary due to the low count of eyes that checked the code of `calrs`.
I figured, I'd contribute some Claude tokens to do a security audit.
**I fully admit I am not an expert in Rust and know very little about frontend technology - I just want to help.**
If you appreciate the contribution, please read on, otherwise feel free to close the PR - no offense taken.

Claude identified a few vulnerabilities (I did a single round, so in my experience there might be more vulnerabilities).
This PR is, in particular, to fix a XSS vulnerability.
Below is Claude's writing on this issue.

If this PR is merged and you so wish, I will open three PRs with the other issues Claude identified:
* [high] OIDC Account Takeover via Automatic Email-Based Account Linking
* [high] Email Header Injection via Unvalidated Guest Name
* [medium] CSRF Cookie Missing `Secure` Flag

---

**Files:**
- `templates/dashboard_event_types.html:61`
- `templates/dashboard_sources.html:28`
- `templates/team_settings.html:219`

**Severity:** High
**Category:** `stored_xss`
**Confidence:** 9/10

### Description

Three dashboard templates embed user-controlled strings directly inside inline JavaScript `onclick` string literals. Minijinja's HTML auto-escaping converts `'` to `&#x27;`, but backslashes (`\`) are not HTML-encoded. The templates wrap the variable with `\'…\'` as a naïve JS-level escape, but this protection can be bypassed: because the attacker controls whether their input begins with `\`, they can cause the template's own backslash to be consumed as an escape for itself, leaving the following apostrophe unprotected.

Vulnerable pattern (identical structure in all three files):

```html
onclick="if(confirm('Delete event type \'{{ et.title }}\'? …')) …"
```

### Exploit Scenario

An authenticated user sets a resource name to `\'; alert(document.cookie)//`. Minijinja HTML-encodes the apostrophe but not the backslash:

```html
onclick="if(confirm('Delete event type \'\\&#x27;; alert(document.cookie)//\'? …'))"
```

The browser decodes HTML entities in attribute values before passing to the JavaScript engine, producing:

```javascript
if(confirm('Delete event type \'\\'; alert(document.cookie)//\'? …'))
```

JavaScript parsing:
1. `\'` — escaped apostrophe, string continues
2. `\\` — escaped backslash, produces a literal `\` in the string value; this escape is now fully consumed
3. `'` — **closes the string** (the `\\` above consumed the backslash, leaving this quote unescaped)
4. `; alert(document.cookie)//` — **executes as JavaScript**

For `dashboard_event_types.html` and `team_settings.html`, team event types and team settings are visible to all team members/admins, making this a true multi-victim stored XSS. For `dashboard_sources.html`, exploitation also impacts any admin who views the page via impersonation.

### Recommendation

Use `data-*` attributes to pass values to JavaScript instead of inline string literals:

```html
<button data-name="{{ et.title }}" data-form-id="del-{{ et.id }}"
        class="delete-btn danger">Delete</button>
<script>
document.querySelectorAll('.delete-btn').forEach(btn => {
    btn.addEventListener('click', () => {
        if (confirm('Delete "' + btn.dataset.name + '"? This cannot be undone.')) {
            document.getElementById(btn.dataset.formId).submit();
        }
    });
});
</script>
```

This eliminates all HTML/JS context confusion. Alternatively, adopt a Content Security Policy that forbids inline event handlers (`script-src 'self'`).

### DIFF
--- a/templates/dashboard_event_types.html
+++ b/templates/dashboard_event_types.html
@@ -58,7 +58,7 @@
             <a href="{{ et.view_url }}">View public page</a>
             {% endif %}
             {% if et.can_manage and et.active_bookings == 0 %}
-            <button type="button" class="danger" onclick="if(confirm('Delete event type \'{{ et.title }}\'? This cannot be undone.')) this.nextElementSibling.submit();">Delete</button>
+            <button type="button" class="danger" data-confirm="Delete event type '{{ et.title }}'? This cannot be undone." onclick="if(confirm(this.dataset.confirm)) this.nextElementSibling.submit();">Delete</button>
             <form method="post" action="{{ et.delete_url }}" style="display: none;"></form>
             {% endif %}
           </div>
--- a/templates/dashboard_sources.html
+++ b/templates/dashboard_sources.html
@@ -25,7 +25,7 @@
           <form method="post" action="/dashboard/sources/{{ s.id }}/test" style="margin: 0;">
             <button type="submit" class="slot-btn" style="font-size: 0.8rem; cursor: pointer;">Test</button>
           </form>
-          <button type="button" class="slot-btn" style="font-size: 0.8rem; color: var(--error-text); cursor: pointer;" onclick="if(confirm('Remove source \'{{ s.name }}\'? This will delete all synced events from this source.')) this.closest('div').querySelector('.remove-form').submit();">Remove</button>
+          <button type="button" class="slot-btn" style="font-size: 0.8rem; color: var(--error-text); cursor: pointer;" data-confirm="Remove source '{{ s.name }}'? This will delete all synced events from this source." onclick="if(confirm(this.dataset.confirm)) this.closest('div').querySelector('.remove-form').submit();">Remove</button>
           <form method="post" action="/dashboard/sources/{{ s.id }}/remove" class="remove-form" style="display: none;"></form>
         </div>
       </div>
--- a/templates/team_settings.html
+++ b/templates/team_settings.html
@@ -216,7 +216,7 @@
 <div class="card" style="margin-top: 2rem; border: 1px solid var(--error-text, #dc2626);">
   <h2 style="font-size: 1rem; margin-bottom: 0.75rem; color: var(--error-text);">Danger zone</h2>
   <p style="color: var(--text-muted); font-size: 0.85rem; margin-bottom: 0.75rem;">Permanently delete this team. Event types will be unlinked (not deleted). This cannot be undone.</p>
-  <button type="button" class="slot-btn" style="font-size: 0.85rem; color: var(--error-text); border-color: var(--error-text); cursor: pointer;" onclick="if(confirm('Delete team \'{{ team_name }}\'? This cannot be undone.')) this.nextElementSibling.submit();">Delete this team</button>
+  <button type="button" class="slot-btn" style="font-size: 0.85rem; color: var(--error-text); border-color: var(--error-text); cursor: pointer;" data-confirm="Delete team '{{ team_name }}'? This cannot be undone." onclick="if(confirm(this.dataset.confirm)) this.nextElementSibling.submit();">Delete this team</button>
   <form method="post" action="/dashboard/teams/{{ team_id }}/delete" style="display: none;"></form>
 </div>
 


olivierlambert:
Thanks @marcotama — much appreciated. I rendered the three templates against the payload in your writeup and a few variants to double-check, and wanted to flag one nuance before merging:

The payload you quoted (`\'; alert(document.cookie)//`, single backslash) actually doesn't pop — MiniJinja passes `\` through unchanged, so the attacker's single backslash becomes a single backslash in the output. With the template's wrapper `\'{{ var }}\'`, this produces `\'\'; alert()//\'` which JavaScript reads as two `\'` escape pairs (string stays open) and the injected `alert` ends up as plain text in the confirm dialog.

The real-world payload needs an **even** number of backslashes before the quote so the quote ends up unescaped. Concretely: `\\'));alert(1);//` — the `\\` pairs up into a literal backslash, the following `'` closes the JS string, `))` closes both `confirm(` and `if(`, then `;alert(1);//` runs unconditionally and eats the trailing template cruft.

So the vulnerability is absolutely real and the fix is correct — just wanted you to have an accurate payload in case you keep probing for similar patterns elsewhere. Merging now and will add a small regression test on main so this class of bug stays caught.

Also yes — please do open the other three. OIDC email-based linking and email header injection both match patterns I'd want verified; CSRF `Secure` flag is quick to settle either way.

olivierlambert:
Closing the loop on the three follow-up findings from your audit, @marcotama — all landed on \`main\`. Thanks again, this was a genuinely useful security round.

### 1. OIDC account takeover via email-based linking — **fixed**

Commit [dbb2712](https://github.com/olivierlambert/calrs/commit/dbb2712).

The bug was real. \`find_or_create_oidc_user\` fell through to matching by email without checking the ID token's \`email_verified\` claim, so any IdP that allows unverified registrations (Keycloak's default) let an attacker attach their \`oidc_subject\` to any existing local account by registering at the IdP with the target's email.

Fix: \`oidc_callback\` now extracts \`email_verified\` (missing claim treated as false — conservative) and passes it to \`find_or_create_oidc_user\`, which gates the email-link and auto-register branches. The \`oidc_subject\` match path (step 1) is unchanged so returning users keep working. Six new tests cover the attack scenario (seed a local admin, run the attack, assert auth_provider/oidc_subject untouched), the legitimate verified flow, the auto-register squat variant, and the returning-user bypass.

### 2. Email header injection via unvalidated guest name — **verified safe, regression-guarded**

Commit [5c2a2a5](https://github.com/olivierlambert/calrs/commit/5c2a2a5).

This one turned out to be a false positive — but worth anchoring with a test. Every email-sending path builds addresses via \`format!(\"{name} <{email}>\").parse::<Mailbox>()\` and subjects via \`Message::builder().subject(…)\`. I ran four realistic CRLF payloads through both APIs:

- \`Mailbox::from_str\` **rejects** CRLF/LF in display names with \"Invalid input\" (the \`?\` propagates through and the email never sends)
- \`.subject(…)\` RFC 2047-encodes any control chars, so \`Subject: Normal\\r\\nBcc: evil\` ends up on the wire as \`Subject: =?utf-8?b?Tm9ybWFsDQpCY2M6?= evil@attacker.com\` — the CRLF is inside the base64 and can't break out to a new header line.

No production fix needed, but lettre is load-bearing for both guarantees. Added two tests that pin the behavior — if a future upgrade relaxes either defense we'll find out at build time, not at exploit time.

### 3. CSRF cookie missing \`Secure\` flag — **fixed**

Commit [a1a7fd0](https://github.com/olivierlambert/calrs/commit/a1a7fd0).

One-line fix to add \`Secure\` to \`csrf_cookie_value\`. \`HttpOnly\` intentionally stays off — the double-submit pattern needs the client JS in \`base.html\` to read the cookie. Added a test asserting the cookie string has \`Secure; SameSite=Lax\` and does NOT have \`HttpOnly\`, so the three invariants are locked in together.

---

Also worth noting in your audit writeup for next time: the XSS payload you quoted in #43 (single backslash, \`\\'; alert(…)//\`) doesn't actually pop — MiniJinja passes \`\\\` through unchanged, so the attacker ends up with \`\\'\\'; alert(…)//\\'\` which JavaScript reads as two \`\\'\` escape pairs (string stays open). The real exploit needs an **even** number of backslashes so they pair up and leave the \`'\` unescaped (\`\\\\')\`), plus a way to close the enclosing \`confirm(\` and \`if(\` without a syntax error — e.g. \`\\\\')); alert(1); //\`. Doesn't change the severity at all — it's still a real stored XSS — but worth knowing if you keep probing for similar patterns elsewhere.

Considering all four issues from this audit closed. Please do open the next round whenever you feel like it.

PR #49, opened 2026-04-23
fix: add required DTSTAMP property to generated ICS calendar invites

Handfish:
Generated .ics calendar attachments were missing the DTSTAMP property, which is required by RFC 5545 for VEVENT components. Some calendar clients (e.g., RustiCal) may reject or mishandle invites without this field. This fix adds DTSTAMP set to the current UTC time in both generate_ics() and generate_cancel_ics().


olivierlambert:
Thanks @Handfish — great catch! DTSTAMP is required by RFC 5545 §3.6.1 and RustiCal is just the strict client that surfaced the bug. Merging now; I'll follow up on `main` with a small regression test so we don't lose this field again.

PR #52, opened 2026-04-24
feat: SMTP config via environment variables

ipkstef:
Adds `CALRS_SMTP_*` env vars as an alternative to `calrs config smtp`. Useful for Docker/container deployments where you want config in the environment rather than the database.

Required: `CALRS_SMTP_HOST`, `CALRS_SMTP_USERNAME`, `CALRS_SMTP_PASSWORD`, `CALRS_SMTP_FROM_EMAIL`
Optional: `CALRS_SMTP_PORT` (default 587), `CALRS_SMTP_TLS_MODE` (`starttls` or `tls`), `CALRS_SMTP_FROM_NAME`

Env vars take priority over the database. Partial config (any var set but required fields missing) errors loudly rather than silently falling through.

Also fixes a hang when using port 465 — `send_email` was unconditionally calling `starttls_relay()` which doesn't work with implicit TLS. Added `SmtpTlsMode` enum and branch on `relay()` vs `starttls_relay()` accordingly.

Admin panel shows env-sourced config status and any misconfiguration errors inline.

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -187,6 +187,33 @@ pub struct ImpersonationInfo {
     pub target_name: String,
 }
 
+/// Validate the session cookie and resolve admin impersonation in one place,
+/// so AuthUser/OptionalAuthUser stay in sync. Returns None when there is no
+/// valid session.
+async fn resolve_session_user(
+    pool: &SqlitePool,
+    jar: &CookieJar,
+) -> Option<(User, Option<ImpersonationInfo>)> {
+    let token = jar.get(SESSION_COOKIE).map(|c| c.value().to_string())?;
+    let real_user = validate_session(pool, &token).await?;
+
+    if real_user.role == "admin" {
+        if let Some(target_id) = jar.get(IMPERSONATE_COOKIE).map(|c| c.value().to_string()) {
+            if target_id != real_user.id {
+                if let Some(target_user) = get_user_by_id(pool, &target_id).await {
+                    let info = ImpersonationInfo {
+                        admin_name: real_user.name.clone(),
+                        target_name: target_user.name.clone(),
+                    };
+                    return Some((target_user, Some(info)));
+                }
+            }
+        }
+    }
+
+    Some((real_user, None))
+}
+
 /// Extractor that requires an authenticated user. Redirects to /auth/login if not authenticated.
 /// Supports admin impersonation: if the `calrs_impersonate` cookie is set and the real user is
 /// an admin, returns the impersonated user instead.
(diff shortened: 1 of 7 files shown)


olivierlambert (commented):
Thanks a lot for this contribution! 🙏 The SMTP-env-var work itself is really nice — clean validation, good error surfacing, well-isolated `SmtpTlsMode` enum, and especially the port-465 hang fix which is a real bug catch. The test coverage on that part is excellent.

Before merging I wanted to share a few thoughts — none of these are blockers, just things worth discussing.

## Scope of the PR

The biggest thing on my mind is that the PR title and description are about SMTP, but the diff actually contains four commits doing fairly different work:

1. `12e0504` — SMTP env vars + TLS mode (matches the title 👍)
2. `e42d756` — Tighten team event-type permissions, add `OptionalAuthUser`, gate booking on slot pages
3. `fb4934f` — Loosen invite gating for `internal` events, show non-public events on team profile
4. `d9bb743` — "View page" vs "View public page" label tweak

Commit 2 is functionally a privilege-escalation fix (regular team members could create/edit/toggle/delete team event types, plus personal-route leakage where ownership was scoped only to `account_id`). It's a meaningful security change and I think it really deserves its own PR with its own description, so reviewers can focus on the auth model in isolation. Commit 3 also partially walks back commit 2 for `internal` events specifically, which is much easier to follow as a focused PR than as a footnote inside the SMTP feature.

Would you be open to splitting this into two PRs? Roughly:
- **PR A**: SMTP env vars + TLS mode + port-465 fix (commit 1, basically already shippable)
- **PR B**: Team-permissions tightening + invite/visibility adjustments (commits 2–4) with a couple of tests around the new `can_manage_event_type` policy

Happy to help review B once it's split out.

## Test coverage

The SMTP commit is well-tested. The auth/permissions rework adds `can_manage_event_type`, `find_manageable_event_type_by_slug`, `OptionalAuthUser`, the `can_book` matrix in `show_group_slots`, and a new `delete_invite` SQL policy — but currently with no tests. Given that it's security-sensitive code, even a couple of unit tests around `can_manage_event_type` (personal owner / team admin / team member / global admin / unrelated user) would give a lot of confidence.

## Smaller technical notes

1. **The port-465 fix is incomplete for DB-stored SMTP config.** In `load_smtp_config`, when reading from the DB, `tls_mode` is hardcoded to `SmtpTlsMode::StartTls`. Anyone who configured port 465 via `calrs config smtp` will still hang — the fix only helps env-var users. Two options: add a `tls_mode` column to `smtp_config` (with a migration), or just document that DB-configured SMTP is STARTTLS-only and recommend env vars for implicit TLS. Either is fine, just worth being explicit.

2. **`delete_invite` now lets the original creator delete even after they leave the team.** The new SQL adds `bi.created_by_user_id = ?` as an OR branch. If a user creates invites and is later removed from the team, they can still delete those invites. Probably intentional, but the inverse — current team admins can't clean up an ex-member's invites without global admin — might be more surprising. Worth confirming the desired behavior.

3. **`find_manageable_event_type_by_slug` uses `UNION ALL ... LIMIT 1`.** Branch order isn't guaranteed by the SQL standard. SQLite tends to preserve it in practice, but if a user has both a personal and a team event type sharing the same slug, the resolved row would be implementation-dependent. An explicit `ORDER BY` would make it deterministic.

4. **Minor:** `{% if et.visibility is defined and et.visibility != 'public' %}` in `team_profile.html` — `is defined` is now belt-and-suspenders since both query branches always return visibility. Can drop it.

Thanks again for the contribution and the careful work on the SMTP piece — the test setup with `SmtpEnvGuard` to serialize env mutation is a nice touch. Happy to chat through any of the above! 🙂

ipkstef:
Hey sorry for the confusion let me fix these PR's i have my local branch running both patches and i cut the pr from main instead of branching. Give me a few mins and i'll fix these pr's

ipkstef:
Re-opened on a clean branch as #56 — same content (commit 12e0504), just isolated from later work on my fork's main that was unintentionally tracked here.

PR #55, opened 2026-04-25
fix: separate team event management from read-only availability viewing

### LINKED ISSUE #53: Fix team event permissions and allow team members to view private team availability
I’m planning to prototype this in my fork and open a PR, but wanted to capture the permission model first in case maintainers prefer a different behavior.

---

## Summary

  Team event permissions currently blur two separate concepts:

  1. Regular team members can modify team event types.
  2. Regular team members cannot view private/internal team event availability without generating or using an invite link.

  This creates a confusing model. A team member can mutate a team event type from the dashboard, but cannot use that same event type as a read-only “when is our team free?” calendar
  unless they go through invite flow.

  ## Current Behavior

  - A user with `team_members.role = 'member'` can manage team event types in several paths:
    - create
    - edit
    - enable/disable
    - delete
    - availability overrides
    - private invite management

  - A logged-in team member visiting:

    `/team/{team_slug}/{private_event_slug}`

    gets blocked with the same invite-token requirement as an anonymous guest.

  - Private/internal event types require `booking_invites` tokens in `show_group_slots`, with no authenticated team-member exception.

  ## Expected Behavior

  Separate management access from read-only availability access:

  - Team admins and global admins can manage team event types.
  - Regular team members cannot mutate team event type configuration.
  - Regular team members can view availability for their team’s private/internal event types without an invite token.
  - Booking private/internal event types remains invite-gated.
  - Anonymous users and non-members still require valid invite tokens.

  ## Desired Access Rules

  - Public event on public team: anyone can view/book.
  - Public event on private team: valid team invite or logged-in team member can view/book.
  - Private/internal team event with valid booking invite: view/book.
  - Private/internal team event as logged-in team member without invite: view only.
  - Private/internal team event as anonymous/non-member without invite: denied.

  ## Implementation Notes

  ### Relevant Areas

  - `src/auth.rs`
    - Existing `AuthUser` extractor
    - Add optional authenticated user helper/extractor

  - `src/web/mod.rs`
    - `dashboard_event_types`
    - `new_event_type_form`
    - `create_event_type`
    - group/team event type handlers
    - overrides handlers
    - invite management handlers
    - `show_group_slots`

  - `templates/slots.html`
    - Render bookable vs read-only slot pills

  ### Acceptance Criteria

  - Regular team member can open a private/internal team event slot page and see availability without `?invite=`.
  - Slot pills are not clickable/bookable for member-only private/internal viewing.
  - Direct `/book` access without a valid invite remains blocked for private/internal event types.
  - Regular team member cannot create, edit, toggle, delete, override, or manage private invites for team event types.
  - Team admin and global admin retain management access.
  - Anonymous/non-member users still need valid invite tokens.
  - Existing internal quick-link behavior remains unchanged.

ipkstef:
Closes #53

Team event permissions currently let regular members mutate event types but block them from viewing their own team's private/internal availability without an invite. This separates management from read-only viewing.

**Management** (create/edit/toggle/delete/overrides/private invites) now requires team admin or global admin across both the unified and legacy group routes. Personal-route handlers also gain `et.team_id IS NULL` so a member who originally created a team event can't bypass via `/dashboard/event-types/{slug}/edit`.

**Read-only viewing** (`show_group_slots`): a logged-in team member sees availability for their team's private/internal events without an invite. Slot pills render as non-clickable spans (`can_book = false`); `/book` and the booking POST stay invite-gated for private/internal. Public events on private teams are bookable by logged-in members without the team invite token. Anonymous users continue to need invite tokens.

**Internal event invite issuance** (named guest + quick link) stays open to any authenticated user, preserving the "any colleague can distribute" model. The team profile page lists private/internal events to logged-in members with a visibility badge so they're discoverable.

Policy is centralized in three helpers so the rules live in one place: `auth.rs::resolve_session_user` (shared by `AuthUser` and a new `OptionalAuthUser`), `web::can_manage_event_type` (by event_type_id), and `web::find_manageable_event_type_by_slug`.

`cargo fmt --check`, `cargo clippy -- -D warnings`, and `cargo test` (545 passing) all clean. No migration; no schema change.

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -404,6 +404,33 @@ pub struct ImpersonationInfo {
     pub target_name: String,
 }
 
+/// Validate the session cookie and resolve admin impersonation in one place,
+/// so AuthUser/OptionalAuthUser stay in sync. Returns None when there is no
+/// valid session.
+async fn resolve_session_user(
+    pool: &SqlitePool,
+    jar: &CookieJar,
+) -> Option<(User, Option<ImpersonationInfo>)> {
+    let token = jar.get(SESSION_COOKIE).map(|c| c.value().to_string())?;
+    let real_user = validate_session(pool, &token).await?;
+
+    if real_user.role == "admin" {
+        if let Some(target_id) = jar.get(IMPERSONATE_COOKIE).map(|c| c.value().to_string()) {
+            if target_id != real_user.id {
+                if let Some(target_user) = get_user_by_id(pool, &target_id).await {
+                    let info = ImpersonationInfo {
+                        admin_name: real_user.name.clone(),
+                        target_name: target_user.name.clone(),
+                    };
+                    return Some((target_user, Some(info)));
+                }
+            }
+        }
+    }
+
+    Some((real_user, None))
+}
+
 /// Extractor that requires an authenticated user. Redirects to /auth/login if not authenticated.
 /// Supports admin impersonation: if the `__Host-calrs_impersonate` cookie is set and the real user is
 /// an admin, returns the impersonated user instead.
@@ -420,39 +447,39 @@ impl FromRequestParts<Arc<AppState>> for AuthUser {
         state: &Arc<AppState>,
     ) -> Result<Self, Self::Rejection> {
         let jar = CookieJar::from_headers(&parts.headers);
-        let token = jar.get(SESSION_COOKIE).map(|c| c.value().to_string());
-
-        let real_user = match token {
-            Some(ref token) => validate_session(&state.pool, token).await,
-            None => None,
-        };
-
-        let real_user = match real_user {
-            Some(u) => u,
-            None => return Err(Redirect::to("/auth/login").into_response()),
-        };
-
-        // Check for impersonation
-        if real_user.role == "admin" {
-            if let Some(target_id) = jar.get(IMPERSONATE_COOKIE).map(|c| c.value().to_string()) {
-                if target_id != real_user.id {
-                    if let Some(target_user) = get_user_by_id(&state.pool, &target_id).await {
-                        let info = ImpersonationInfo {
-                            admin_name: real_user.name.clone(),
-                            target_name: target_user.name.clone(),
-                        };
-                        return Ok(AuthUser {
-                            user: target_user,
-                            impersonation: Some(info),
-                        });
-                    }
-                }
-            }
+        match resolve_session_user(&state.pool, &jar).await {
+            Some((user, impersonation)) => Ok(AuthUser {
+                user,
+                impersonation,
+            }),
+            None => Err(Redirect::to("/auth/login").into_response()),
         }
+    }
+}
 
-        Ok(AuthUser {
-            user: real_user,
-            impersonation: None,
+/// Extractor that does NOT require authentication. Returns Some(user) for
+/// authenticated requests (mirroring AuthUser's impersonation resolution) and
+/// None for anonymous/invalid sessions — never redirects.
+pub struct OptionalAuthUser {
+    pub user: Option<User>,
+    pub impersonation: Option<ImpersonationInfo>,
+}
+
+impl FromRequestParts<Arc<AppState>> for OptionalAuthUser {
+    type Rejection = std::convert::Infallible;
+
+    async fn from_request_parts(
+        parts: &mut Parts,
+        state: &Arc<AppState>,
+    ) -> Result<Self, Self::Rejection> {
+        let jar = CookieJar::from_headers(&parts.headers);
+        let (user, impersonation) = match resolve_session_user(&state.pool, &jar).await {
+            Some((u, info)) => (Some(u), info),
+            None => (None, None),
+        };
+        Ok(OptionalAuthUser {
+            user,
+            impersonation,
         })
     }
 }
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -7672,9 +7672,17 @@ async fn team_profile_page(
         None => return Html("Team not found.".to_string()),
     };
 
-    // Gate private teams behind invite token
+    // Logged-in team members (and global admins) bypass the private-team
+    // invite token gate — their membership is the access. Compute once and
+    // reuse for both the gate below and the event-type filter further down.
+    let viewer_is_member_or_admin = match &optional_auth.user {
+        Some(u) if u.role == "admin" => true,
+        Some(u) => is_team_member(&state.pool, &u.id, &team_id).await,
+        None => false,
+    };
+
     let passed_invite = query.get("invite").cloned().unwrap_or_default();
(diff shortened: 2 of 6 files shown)


olivierlambert (commented):
Thanks so much for splitting this out — really appreciate the responsiveness 🙏 The shape of this PR is much easier to reason about now: management policy is centralized in `can_manage_event_type` / `find_manageable_event_type_by_slug`, session resolution is shared between `AuthUser` and the new `OptionalAuthUser`, and the access matrix in `show_group_slots` reads cleanly. Nice job covering all four entry points (`team_profile_page`, `show_group_slots`, `show_group_book_form`, `handle_group_booking`) — the consistency between the view-side `can_book` flag and the actual booking POST gate is exactly right.

A few thoughts, none blocking. I'd be happy to merge after a chat on the test point at minimum.

## Test coverage

This is the one I'd most like to see addressed before merge, given the security-sensitive nature of the change. The existing 545 tests pass, but none of the new helpers are directly covered:

- `can_manage_event_type` — even a small unit test sweep (global admin / personal owner / team admin / team member / unrelated user, × personal event / team event) would lock in the policy and protect against future regressions. It's a focused enough function that a `#[sqlx::test]`-style test against an in-memory DB is pretty mechanical.
- `find_manageable_event_type_by_slug` — at least one case where a personal and a team event type share a slug (see point 2 below).
- The `can_view` / `can_book` matrix in `show_group_slots` — a couple of integration tests would be ideal but I understand the project doesn't have a web-handler test harness yet (CLAUDE.md flags this as the biggest coverage gap), so I won't insist on those here.

## Smaller technical points

1. **`delete_invite` still lets the original creator delete after leaving the team.** The new SQL adds `bi.created_by_user_id = ?` as an OR branch alongside the team-admin / personal-owner branches. So if a member creates invites as part of the "any colleague can distribute" model, then is removed from the team, they retain delete access to those invites — while current team admins can't clean up unrelated ex-member invites without going through global admin. Probably intentional, but worth confirming. If yes, a one-line comment explaining the rationale would help future-me.

2. **`find_manageable_event_type_by_slug` uses `UNION ALL ... LIMIT 1` without `ORDER BY`.** SQL doesn't guarantee branch order; SQLite tends to preserve it in practice but it's implementation-dependent. If a user has both a personal event and a team event with the same slug (rare but possible), the resolved row would be undefined behavior per the standard. Adding `ORDER BY` (e.g., personal-first explicitly, or ordering by team_id NULLS FIRST) would make this deterministic.

3. **`et.visibility is defined`** in `team_profile.html` is now belt-and-suspenders since both query branches in `team_profile_page` always return visibility. Can drop the `is defined` check.

4. **Tiny perf nit** (genuinely minor, ignore if you prefer): `team_id_for_access` re-queries `SELECT team_id FROM event_types WHERE id = ?` after the initial event-type lookup already had the team's identity available via the JOIN. Could be threaded through from the first query. Not worth a re-roll on its own.

## Things I particularly liked

- Threading `optional_auth` through all four endpoints (view, slot list, book form, booking POST) so the policy is consistent end-to-end — that's better than just patching the view.
- Pre-computing `viewer_is_member_or_admin` once per handler instead of recomputing inside conditionals.
- The "preserve legacy 'Event type not found' to avoid leaking existence" comment in the new `can_view` branch — good security hygiene.
- The clear access-matrix comment block (`public team + public event → anyone`, etc.) right next to the `match` — really helpful for future readers.

Thanks again for the careful work and for being patient with the back-and-forth! 🙂

olivierlambert:
Friendly ping @ipkstef, no rush at all 🙂 Whenever you have time, this one needs a rebase against main (currently conflicting) plus a look at the test coverage point from the review. Happy to merge once those land.

olivierlambert (changes requested):
Flipping to Changes Requested so the dashboard state is accurate. The two outstanding items from earlier are:

1. Rebase on current main (the branch is conflicting today).
2. The test coverage point from the review above (manageability paths through `can_manage_event_type` need a regression test).

Take your time. If you'd rather hand this off, the offer to pick it up on a follow-up branch still stands 🙏

ipkstef:
Apologies for the delay on this one. Just pushed a refreshed branch:

- Rebased onto current `main` and resolved conflicts.
- Added the regression tests you asked for around `can_manage_event_type` — global admin can manage personal + team event types, personal owner can manage their own, unrelated user can't manage another user's personal or team event type, team admin can manage team event types, team member cannot.
- Added a deterministic-ordering test for `find_manageable_event_type_by_slug` so personal events win over team events on a slug collision (now wrapped in a subquery with `ORDER BY (team_id IS NULL) DESC`).
- Also tightened `delete_invite` so authenticated colleagues who issued an internal-event invite can no longer delete it — only admin/owner/team-admin per `can_manage_event_type`. Template hides the Delete button when the viewer can't delete.

fmt/clippy/tests all green.

olivierlambert (approved):
🎉 Welcome back @ipkstef, this is great. Both original asks are squarely addressed:

- ✅ Clean rebase on current main (merge-base at `2e4ab5f`, no conflicts).
- ✅ 8 new regression tests at `src/web/mod.rs:16148-16377` covering the full matrix I asked for and then some: global admin yes on personal+team, personal owner yes / unrelated user no, team admin yes / member no / outsider no, team-admin-elsewhere can't reach unrelated personal events, slug-collision determinism (personal wins over team), and team-member (non-admin) doesn't resolve via `find_manageable_event_type_by_slug`. Thorough.

The branch grew beyond the original two-item ask into a broader pass on team/invite access control consistency. After looking through each commit, every one is genuinely on-theme (who can manage / view / book a team event), the gates use `can_manage_event_type` consistently, and the new behaviour is coherent. Worth landing as a single PR rather than splitting. Two specific things worth calling out for changelog and user awareness, neither blocking:

1. **`delete_invite` is now stricter** (`src/web/mod.rs:6347-6400`): a non-admin team member who legitimately created an internal-event invite via the "any authenticated user" path can no longer delete that invite. Only owners, team admins, and global admins can. The `can_delete_invites` flag passed to the template makes the UI match. Acknowledged in your commit message; worth a sentence in the release notes so anyone relying on the older "creator can delete" behaviour knows to either let invites expire or hit max-uses.

2. **`team_profile_page` now surfaces private/internal event-type titles and slugs to all logged-in team members** (`src/web/mod.rs:7713-7736`), where previously only public ones were listed. Booking is still invite-gated, but the slug becomes guessable for any member. This is consistent with `show_group_slots` letting members view (not book) any team event, so it's the right direction; just noting it's a small widening within the team boundary.

Authorization paths look clean: CSRF verified on every mutating handler I checked, SQL parameterised throughout, no new `.unwrap()` introduced.

`cargo test`: 663/663 on a clean tree, `cargo fmt`: clean, `cargo clippy`: clean.

Approving and ready to merge. Really nice work on the test coverage in particular 🙏

PR #56, opened 2026-04-25
feat: SMTP config via environment variables

ipkstef:
Adds `CALRS_SMTP_*` env vars as an alternative to `calrs config smtp`. Useful for Docker/container deployments where you want config in the environment rather than the database.

Required: `CALRS_SMTP_HOST`, `CALRS_SMTP_USERNAME`, `CALRS_SMTP_PASSWORD`, `CALRS_SMTP_FROM_EMAIL`
Optional: `CALRS_SMTP_PORT` (default 587), `CALRS_SMTP_TLS_MODE` (`starttls` or `tls`), `CALRS_SMTP_FROM_NAME`

Env vars take priority over the database. Partial config (any var set but required fields missing) errors loudly rather than silently falling through.

Also fixes a hang when using port 465 — `send_email` was unconditionally calling `starttls_relay()` which doesn't work with implicit TLS. Added `SmtpTlsMode` enum and branch on `relay()` vs `starttls_relay()` accordingly.

Admin panel shows env-sourced config status and any misconfiguration errors inline.

---
_Re-opened on a clean branch from #52 — same single commit, just isolated from later work on my fork's main._

### DIFF
--- a/migrations/052_smtp_tls_mode.sql
+++ b/migrations/052_smtp_tls_mode.sql
@@ -1,0 +1,5 @@
+-- Per-row TLS mode for the SMTP transport. NULL/'starttls' (default) means
+-- the existing STARTTLS handshake; 'tls' means implicit TLS (typically port
+-- 465), which previously caused send_email() to hang because the code path
+-- unconditionally used starttls_relay().
+ALTER TABLE smtp_config ADD COLUMN tls_mode TEXT NOT NULL DEFAULT 'starttls';
--- a/src/commands/config.rs
+++ b/src/commands/config.rs
@@ -91,6 +91,20 @@ pub async fn run(pool: &SqlitePool, key: &[u8; 32], cmd: ConfigCommands) -> Resu
                     Some(name)
                 }
             });
+            let tls_mode = {
+                let raw = prompt("TLS mode (starttls/tls, default starttls)");
+                let normalized = raw.trim().to_ascii_lowercase();
+                match normalized.as_str() {
+                    "" | "starttls" => "starttls",
+                    "tls" => "tls",
+                    other => {
+                        anyhow::bail!(
+                            "Invalid TLS mode '{}'. Use 'starttls' (default, port 587) or 'tls' (implicit TLS, port 465).",
+                            other
+                        );
+                    }
+                }
+            };
 
             let password_enc = crate::crypto::encrypt_password(key, &password)?;
             let id = Uuid::new_v4().to_string();
@@ -99,8 +113,8 @@ pub async fn run(pool: &SqlitePool, key: &[u8; 32], cmd: ConfigCommands) -> Resu
             sqlx::query("DELETE FROM smtp_config").execute(pool).await?;
 
             sqlx::query(
-                "INSERT INTO smtp_config (id, host, port, username, password_enc, from_email, from_name)
-                 VALUES (?, ?, ?, ?, ?, ?, ?)",
+                "INSERT INTO smtp_config (id, host, port, username, password_enc, from_email, from_name, tls_mode)
+                 VALUES (?, ?, ?, ?, ?, ?, ?, ?)",
             )
             .bind(&id)
             .bind(&host)
@@ -109,10 +123,17 @@ pub async fn run(pool: &SqlitePool, key: &[u8; 32], cmd: ConfigCommands) -> Resu
             .bind(&password_enc)
             .bind(&from_email)
             .bind(&from_name)
+            .bind(tls_mode)
             .execute(pool)
             .await?;
 
-            println!("{} SMTP configured ({}:{})", "✓".green(), host, port);
+            println!(
+                "{} SMTP configured ({}:{}, {})",
+                "✓".green(),
+                host,
+                port,
+                tls_mode
+            );
         }
         ConfigCommands::Show => {
             let smtp: Option<(String, i32, String, String, Option<String>, bool)> = sqlx::query_as(
--- a/src/db.rs
+++ b/src/db.rs
@@ -223,6 +223,10 @@ pub async fn migrate(pool: &SqlitePool) -> Result<()> {
             "051_per_member_frequency_limit",
             include_str!("../migrations/051_per_member_frequency_limit.sql"),
         ),
+        (
+            "052_smtp_tls_mode",
+            include_str!("../migrations/052_smtp_tls_mode.sql"),
+        ),
     ];
 
     let mut applied_count = 0u32;
@@ -804,7 +808,7 @@ mod tests {
             .fetch_one(&pool)
             .await
             .unwrap();
-        assert_eq!(count.0, 51, "All 51 migrations should be tracked");
+        assert_eq!(count.0, 52, "All 52 migrations should be tracked");
     }
 
     #[tokio::test]
@@ -818,7 +822,7 @@ mod tests {
             .fetch_one(&pool)
             .await
             .unwrap();
-        assert_eq!(count.0, 51, "Still 51 migrations after second run");
+        assert_eq!(count.0, 52, "Still 52 migrations after second run");
     }
 
     #[tokio::test]
--- a/src/email.rs
+++ b/src/email.rs
@@ -1,4 +1,4 @@
-use anyhow::Result;
+use anyhow::{bail, Result};
 use chrono::NaiveDateTime;
 use chrono_tz::Tz;
 use fluent_bundle::{FluentArgs, FluentValue};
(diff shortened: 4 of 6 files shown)


ipkstef:
Bringing over the SMTP-relevant callout from @olivierlambert's review on #52 so it doesn't get lost:

> **The port-465 fix is incomplete for DB-stored SMTP config.** In `load_smtp_config`, when reading from the DB, `tls_mode` is hardcoded to `SmtpTlsMode::StartTls`. Anyone who configured port 465 via `calrs config smtp` will still hang — the fix only helps env-var users. Two options: add a `tls_mode` column to `smtp_config` (with a migration), or just document that DB-configured SMTP is STARTTLS-only and recommend env vars for implicit TLS. Either is fine, just worth being explicit.

Slight preference for the migration so the CLI / admin UI stay first-class, but happy to go doc-only if you'd rather keep the schema lean. Let me know which path and I'll add it to this PR.

olivierlambert (commented):
Thanks for the clean re-roll! 🙏 This is exactly the split I was hoping for — the SMTP feature stands on its own and is much easier to review in isolation. Most of my praise from #52 carries over (clean validation, nice `SmtpTlsMode` enum, the `SmtpEnvGuard` Mutex pattern for env-mutating tests, the port-465 hang catch, and good test coverage on the env paths).

One substantive point still applies, which I'd like to see addressed before merging — sorry to re-raise it but I think it matters:

## The port-465 fix is incomplete for DB-stored SMTP config

`load_smtp_config` hardcodes `tls_mode: SmtpTlsMode::StartTls` when reading from the DB. So anyone who configured port 465 via `calrs config smtp` (the existing CLI path) will still hit the original `starttls_relay()` hang — the fix only helps users who switch to env vars.

A few possible paths, any of these would address it:

1. **Add a `tls_mode` column** with a small migration (`045_smtp_tls_mode.sql`), default `'starttls'`, and update the `calrs config smtp` interactive prompt to ask for it. Most thorough.
2. **Infer from port** in `load_smtp_config`: if `port == 465`, use `SmtpTlsMode::Tls`, else `StartTls`. Pragmatic and matches RFC 8314 conventions. Less flexible but covers the actual bug.
3. **Document the limitation explicitly** in CLAUDE.md / README and in the admin panel SMTP section — "DB-configured SMTP is STARTTLS-only; for port 465 use env vars." Cheapest, but leaves a footgun.

I have no strong preference between (1) and (2). (2) is the smallest diff and resolves the hang for existing users without making them touch config; (1) is more honest about what's stored. (3) feels like a step backward when the fix is so close.

## Smaller technical points

These are all carried over from #52 and remain minor:

- **`load_smtp_status` queries without `WHERE enabled = 1`** while `load_smtp_config` queries with it. I think this is intentional (admin should see "configured but disabled"), but a one-line comment would help future readers not "fix" the inconsistency.
- **`smtp_env_error()` test helper** uses `panic!` inside a match — works fine, just slightly unusual. Could use `expect_err()` for the same effect with clearer intent.

Once the port-465 / DB-config gap is addressed (or explicitly documented if you'd rather defer), this is ready to merge from my side. Thanks again for the clean separation! 🙂

olivierlambert:
Let's go with option (1), the `tls_mode` column + migration. Matches your preference and keeps the CLI/admin UI first-class.

A few notes to keep the diff tight:

- Migration number is `045_smtp_tls_mode.sql` (last is `044_booking_claim.sql`). Don't forget to register it in `src/db.rs` in the `migrations` array, otherwise it won't run on existing deployments.
- Default the column to `'starttls'` so existing rows keep working unchanged.
- Update the `calrs config smtp` interactive prompt to ask for the mode (default starttls, accept `tls` for port 465). The admin panel can stay read-only on this for now if it's easier, happy to take that as a follow-up.
- The two smaller nits from the review (comment on the `enabled` filter difference, and `expect_err` over `panic!` in the test helper) would be nice to fold in here too, but no blocker.

Thanks again!

olivierlambert:
Friendly ping @ipkstef, no rush 🙂 See my comment above picking option (1), the `tls_mode` column + migration. Happy to merge once that lands.

wallon-ines:
Is it possible to make SMTP username and password optional? For, e.g, if it uses an SMTP relay.

olivierlambert:
Hey @ipkstef, no pressure, just another gentle ping on this one 🙂 The env-var path itself looks great, the only thing still outstanding from my side is the port-465 / DB-config gap, with option (1) selected: a `045_smtp_tls_mode.sql` migration adding the `tls_mode` column (default `'starttls'`), registered in the `src/db.rs` migrations array, plus updating the `calrs config smtp` interactive prompt to ask for the mode. Both of the smaller nits (the `enabled` filter comment and `expect_err` in the test helper) are still optional, fold them in only if convenient.

Heads up that `main` has moved a few migrations forward since my comment. If you check `migrations/` and `src/db.rs` you'll want to grab the next free number when you push, not necessarily `045` anymore.

@wallon-ines great point on optional username/password for relays. Currently `SmtpConfig.username` and `.password` are `String` (not `Option<String>`) and `send_email` unconditionally calls `Credentials::new(...)` + `.credentials(creds)`, so the relay-without-auth case isn't supported today. The shape of the change is small (`Option<String>` on both fields, gate the `.credentials()` call), but it touches the DB-stored path too, so I'd rather not bolt it onto this PR. Happy to take it as a separate follow-up issue once #56 lands. Want to file one, or shall I?

@ipkstef if you're stretched for time and would rather hand this off, totally fine, just say the word and I'll pick up the TLS-mode column on a follow-up branch with you credited. Otherwise looking forward to seeing the migration land 🙏

olivierlambert (changes requested):
Flipping to Changes Requested so this surfaces in the right bucket. The TLS-mode column + migration from option (1) is still the open ask. No rush 🙂 and the offer to take it over on a follow-up branch still stands if your bandwidth's elsewhere.

ipkstef:
Hey, sorry for the long quiet — I'm back. Just pushed an update addressing the review feedback:

- Rebased onto current `main` (clean rebase).
- Added migration `052_smtp_tls_mode.sql` so the port-465 fix applies to existing DB-configured SMTP users, not just env-var users. New column defaults to `'starttls'`, and `'tls'` selects implicit TLS for port 465.
- `calrs config smtp` interactive prompt now asks for TLS mode.
- Added the comment explaining why `load_smtp_status` omits the `WHERE enabled = 1` filter.
- Swapped the manual `panic!` match in `smtp_env_error()` for `expect_err()` (with a redacted `Debug` impl on `SmtpConfig` so the password doesn't leak in test failure output).

Admin panel kept read-only for TLS mode, and SMTP username/password for unauthenticated relays left as a follow-up per your suggestion. fmt/clippy/tests all green.

olivierlambert (approved):
🎉 Welcome back @ipkstef, and great to see this addressed in such a complete way. Going through everything:

- ✅ Migration `052_smtp_tls_mode.sql` adds `tls_mode TEXT NOT NULL DEFAULT 'starttls'` so existing rows keep STARTTLS behaviour. Registered in `src/db.rs:226-229`. Migration count assertions bumped to 52.
- ✅ `load_smtp_config` now reads the column at `src/email.rs:1711-1722`, with `unwrap_or(StartTls)` as a sane fallback for unparseable values (operators don't get locked out of email sending if someone hand-edits the row).
- ✅ `send_email` correctly branches on `tls_mode` at `src/email.rs:1885-1899` (`relay()` for TLS, `starttls_relay()` for STARTTLS). This is the actual port-465 hang fix for DB-configured SMTP.
- ✅ `calrs config smtp` CLI prompt at `src/commands/config.rs:94-107` asks for the mode with `starttls` as the default and rejects garbage input.
- ✅ Both prior nits folded in: the `enabled`-filter comment at `src/email.rs:1737-1743`, and `expect_err()` in the test helper at `src/email.rs:2511-2515`.
- ✅ Bonus security win: custom `Debug` impl for `SmtpConfig` at `src/email.rs:46-58` redacts the password, so the `expect_err` failure output (and any future debug logs) can't leak credentials. Nice touch.

`cargo test`: 662/662, `cargo fmt`: clean, `cargo clippy`: clean.

One tiny non-blocking observation, easy follow-up if you feel like it: `templates/admin.html:332` changed from "Configure via CLI (`calrs config smtp`)" to "Configure via environment variables: ..." which loses the CLI discovery hint even though both paths are fully supported now. A one-line tweak to mention both options would be a nicer onboarding pointer. Not blocking the approval.

Approving and ready to merge. Thanks for the thorough turnaround 🙏

PR #74, opened 2026-04-29
landing: add lightbox to showcase-grid items

guzu:
## Summary

- Adds a lightbox overlay to all screenshot items in the showcase grid
- Click any screenshot to open a full-size view with caption
- Close via the ✕ button, clicking outside the image, or pressing Escape

### DIFF
--- a/index.html
+++ b/index.html
@@ -229,6 +229,7 @@
       overflow: hidden;
       border: 1px solid var(--border);
       transition: all 0.3s;
+      cursor: zoom-in;
     }
     .showcase-item:hover {
       border-color: rgba(99, 102, 241, 0.3);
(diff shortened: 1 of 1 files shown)


olivierlambert (changes requested):
Thanks a lot for the contribution, @guzu — really nice first PR! 🎉 The lightbox is dependency-free, scoped in an IIFE, restores `body` overflow on close, and reuses existing CSS variables. That's exactly the style this landing page is written in.

A couple of things to address before merge, plus some nice-to-haves.

## Please fix before merge

### 1. ARIA labels are in French on an English page
The page is `lang="en"` and every other string is English, but the new ARIA labels are French:

```html
<div class="lightbox" ... aria-label="Image agrandie">
<button class="lightbox-close" ... aria-label="Fermer">
```

Screen-reader users on the English site will hear French. Could you switch them to e.g. `aria-label="Enlarged screenshot"` and `aria-label="Close"`?

### 2. Keyboard users can't open the lightbox
`.showcase-item` is a plain `<div>` with no `tabindex`, no `role`, and only a `click` listener. The `cursor: zoom-in` advertises it as interactive, but keyboard users have no way to trigger it. Minimum fix:

```js
document.querySelectorAll('.showcase-item').forEach(function (item) {
  item.setAttribute('tabindex', '0');
  item.setAttribute('role', 'button');
  item.addEventListener('click', function () { /* existing */ });
  item.addEventListener('keydown', function (e) {
    if (e.key === 'Enter' || e.key === ' ') {
      e.preventDefault();
      item.click();
    }
  });
});
```

A `:focus-visible` style on `.showcase-item` would also help users see where they are.

## Nice-to-haves (optional)

- **Focus management.** On open, save `document.activeElement` and focus `#lightbox-close`; on close, restore the saved element. Without this, tab order leaks into the page behind the modal and focus is lost on close. A couple of extra lines.
- **Avoid shadowing `window.open` / `window.close`.** Inside the IIFE, `function open(...)` and `function close()` shadow the globals. Renaming to `openLightbox` / `closeLightbox` is clearer.
- **Idiomatic caption read.** `(item.querySelector('.caption') || {}).textContent || ''` works but the file already uses modern syntax elsewhere — `item.querySelector('.caption')?.textContent ?? ''` is shorter.
- **Close-button placement on short viewports.** `top: -2.5rem; right: 0;` sits *above* the inner container. On landscape phones / short windows it can clip against the top edge. A button absolutely positioned inside the image (top-right with a translucent background) is more robust.
- **`aria-describedby`** linking `#lightbox-img` to `#lightbox-caption` so screen readers announce them together.

## Not blocking
- No security concerns (no `innerHTML` from external input, only `textContent` and `src`/`alt` reads from local DOM).
- No tests needed — landing-page HTML/CSS, visual verification only.

Thanks again, looking forward to the next round! 🙏

guzu:
I fixed every comments, blocking (ARIA labels and keyboard accessibility) as well as optional ones.

olivierlambert (approved):
Beautifully done, @guzu — every blocking item and every optional one addressed. Walking through them:

- ✅ ARIA labels in English (`Enlarge screenshot` / `Close`)
- ✅ `.showcase-item` has `tabindex="0"`, `role="button"`, Enter/Space handler, plus a `:focus-visible` outline
- ✅ Focus management: saves `document.activeElement`, focuses the close button on open, restores on close
- ✅ Renamed `open`/`close` → `openLightbox`/`closeLightbox`
- ✅ `?.textContent ?? ''` for the caption
- ✅ Close button moved inside the inner container with a translucent circular background — way more robust on short viewports
- ✅ `aria-describedby="lightbox-caption"` on the image

Thanks for the thorough follow-through, this is exactly the kind of contribution we love to see. Merging shortly. 🚀

Issue #54, opened 2026-04-25
Feature: Add localization management


gwen-lg:
Do you plan to add localization/translation management in the near future?

I know some peoples than can be interested to use Cal.rs, but they need an interface (at least for the appointment booking part) in french.

Thank you.

olivierlambert:
Hi! Done. See https://github.com/olivierlambert/calrs#localization-1 for more details.

gwen-lg:
@olivierlambert  Thank you

gwen-lg:
After testing, on the different appointment booking pages the translation is always based on the browser language, even if a specific language has been configured in the profile.

PR #76, opened 2026-04-30
fix translation management on the different appointment booking pages

gwen-lg:
related to issue #54 

this add translation management from user profile for the different appointment booking pages :
- user_profile
- show_slots_for_user
- show_book_form_for_user
- handle_booking_for_user

### DIFF
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -7484,7 +7484,7 @@ async fn user_profile(
         Some(u) => u,
         None => return Html("User not found.".to_string()),
     };
-    let lang = language.unwrap_or(crate::i18n::detect_from_headers(&headers).into());
+    let lang = crate::i18n::resolve(language.as_deref(), &headers);
 
     let event_types: Vec<(String, String, Option<String>, i32)> = sqlx::query_as(
         "SELECT et.slug, et.title, et.description, et.duration_min
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -8217,15 +8217,35 @@ async fn show_slots_for_user(
     if username.contains('+') {
         return show_dynamic_group_slots(&state, &headers, &username, &slug, &query).await;
     }
-    let lang = crate::i18n::detect_from_headers(&headers);
-    let et: Option<(String, String, String, Option<String>, i32, i32, i32, i32, String, Option<String>, String, String, Option<String>, Option<String>, String, String)> = sqlx::query_as(
-        "SELECT et.id, et.slug, et.title, et.description, et.duration_min, et.buffer_before, et.buffer_after, et.min_notice_min, et.location_type, et.location_value, u.id, u.name, u.title, u.avatar_path, et.visibility, et.default_calendar_view
-         FROM event_types et
-         JOIN accounts a ON a.id = et.account_id
-         JOIN users u ON u.id = a.user_id
-         WHERE u.username = ? AND et.slug = ? AND et.enabled = 1 AND u.enabled = 1",
+
+    let user: Option<(
+        String,
+        String,
+        Option<String>,
+        Option<String>,
+        Option<String>,
+    )> = sqlx::query_as(
+        "SELECT id, name, title, avatar_path, language FROM users WHERE username = ? AND enabled = 1",
     )
     .bind(&username)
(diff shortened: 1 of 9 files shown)


src/web/mod.rs
```diff
@@ -8414,19 +8418,39 @@ async fn show_slots_for_user(
...
-         WHERE u.username = ? AND et.slug = ? AND et.enabled = 1 AND u.enabled = 1",
+         WHERE a.user_id = ? AND et.slug = ? AND et.enabled = 1",
     )
-    .bind(&username)
+    .bind(&host_user_id)
     .bind(&slug)
     .fetch_optional(&state.pool)
     .await
-    .unwrap_or(None);
+    .expect("EventTypes query failed");
```
olivierlambert: Blocking: this is a regression. The previous code used `.unwrap_or(None)` and `CLAUDE.md` calls out that all web handlers must be crash-proof (no `.unwrap()` / `.expect()`). A transient SQLite error here would now panic the request task instead of returning a soft "Event type not found" page.

```suggestion
    .unwrap_or(None);
```

src/web/mod.rs
```diff
@@ -8414,19 +8418,39 @@ async fn show_slots_for_user(
...
+        "SELECT id, name, title, avatar_path, language FROM users WHERE username = ? AND enabled = 1",
+    )
+    .bind(&username)
+    .fetch_optional(&state.pool)
+    .await
+    .unwrap_or(None);
+
+    let (host_user_id, host_name, host_title, host_avatar_path, user_lang) = match user {
+        Some(user) => user,
+        None => return Html("User type not found.".to_string()),
```
olivierlambert: Tiny typo: this branch fires when the **user** lookup fails, not the event type. Should read "User not found." (the "Event type not found." message a few lines below is the right one for the next match).

```suggestion
        None => return Html("User not found.".to_string()),
```

src/web/mod.rs
```diff
@@ -8414,19 +8418,39 @@ async fn show_slots_for_user(
...
+    };
+
+    let lang = user_lang.unwrap_or(crate::i18n::detect_from_headers(&headers).into());
```
olivierlambert: Two suggestions for this `unwrap_or` (same pattern is repeated in `user_profile`, `show_book_form_for_user`, `handle_booking_for_user`):

1. `unwrap_or(...)` evaluates the fallback eagerly, so `detect_from_headers(&headers)` runs even when the user has a language set. Use `unwrap_or_else(|| ...)` for lazy evaluation.
2. There's already a helper for exactly this in `src/i18n.rs` — `crate::i18n::resolve(user_lang.as_deref(), &headers)` — used by dashboard handlers. If the signature fits, switching all four sites to it would deduplicate the logic.

Not blocking, just a polish pass.

olivierlambert (changes requested):
Thanks a lot for tackling this, Gwen! Pulling the host's language preference into the booking flow is a nice improvement, and the translations read well. Two small blockers and a few notes inline.

**Behavior change worth confirming.** The booking pages now display in the **host's** preferred language whenever it's set, falling back to the guest's `Accept-Language` header otherwise. Previously these were always guest-language. That's a defensible UX (host-branded booking page) and matches the PR title, but it's a real semantic change — a French-speaking guest visiting an Italian host's page will now see Italian. Worth a one-liner in the PR description / release notes so it's not surprising.

**Branch target.** Per `CLAUDE.md`, new translatable strings should land on the long-lived `i18n` branch first (Weblate syncs from there) rather than `main`. Could you retarget the PR to `i18n`? That gives translators time to fill in keys before the next merge into `main`.

**Estonian (`et`).** No new keys added — that's fine, stub locales fall back to English at runtime per `CLAUDE.md`. Just flagging for awareness.

Thanks again for the contribution!

src/web/mod.rs
gwen-lg: updated, a test that remained by mistake

src/web/mod.rs
gwen-lg: fixed

src/web/mod.rs
gwen-lg: Okay, I hadn't checked if the feature existed, I've fix it.

PR #77, opened 2026-05-04
Fix(auth): prevent user enumeration via login timing oracle

marcotama:
Hi again, `calrs` team!

Thanks for looking into my previous PR!
I have finally gotten around to making another round of security auditing with Claude.
This round seems to have found fewer "high" security risks.

This PR in particular mitigates the highest security risk Claude identified (and fixed).

When at login, we want a bad actor "making up" credentials not to know whether the username was wrong OR the password was wrong.
However, currently, if the username is wrong, the result is returned immediately, vs if the password is wrong, the result is returned after a moment, the time necessary to run Argon2. This difference allows the attacker to tell the two scenarios apart.

This fix basically always runs Argon2 regardless, using a dummy hash if the user was not found in the DB.

---

The *medium* issues Claude found are these:
* Use of `thread_rng` instead of `OsRng`; Claude claims that OsRng is the appropriate standard, as the other lives in userspace.
* The time between the check for any registered user and the time of storing the new first-admin account, allows multiple admin accounts to be created.
* > The CSRF token verification uses standard string equality, which is not guaranteed to be constant-time. This creates a theoretical timing oracle for CSRF token forgery.
* The rate-limiter uses the first IP in `X-Forwarded-For`, but it should use the "rightmost non-trusted IP" instead, as that's not spoofable.
* > Raw database errors, template rendering errors, and OIDC token exchange errors are returned directly to HTTP clients
* The admin-configurable `company_link` can be maliciously set to `javascript:` - the schema should be checked to be `http`/`https`.
* The OICD `client_secret` is stored in plaintext.

For the full list plus more details and suggested remediations, I am attaching the full report: 
[SECURITY-REPORT.md](https://github.com/user-attachments/files/27326622/SECURITY-REPORT.md)

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -41,6 +41,10 @@ pub fn verify_password(password: &str, hash: &str) -> bool {
         .is_ok()
 }
 
+// Pre-computed Argon2id hash used as a timing dummy when no user is found,
+// preventing user enumeration via response-time differences.
+const DUMMY_HASH: &str = "$argon2id$v=19$m=19456,t=2,p=1$AAAAAAAAAAAAAAAAAAAAAA$AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA";
+
 // --- Session management ---
 
 fn generate_session_token() -> String {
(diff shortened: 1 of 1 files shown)


olivierlambert:
Thanks a lot for the report and the fix, @marcotama! I went through it carefully:

- `DUMMY_HASH` parses cleanly as Argon2id with the default parameters (`m=19456, t=2, p=1`), so `verify_password()` runs to completion and doesn't short-circuit at PHC parsing.
- Timing on `argon2 0.5.3` (median of 20 warm iterations): dummy verify ≈ 10.7ms, real-wrong-password ≈ 10.4ms, real-correct-password ≈ 10.3ms. Within noise.
- Behavior preserved for all three login outcomes, including the `password_hash IS NULL` case (it falls through the gated `Some(u) if password_ok` arm to the same error).

Merging now. I'll open issues for the remaining medium findings in the report so they don't get lost.

PR #89, opened 2026-05-04
fix(auth): grant first-admin role atomically in INSERT

### LINKED ISSUE #79: auth: race condition when granting first-admin role on registration
From the [security audit](https://github.com/user-attachments/files/27326622/SECURITY-REPORT.md) (FINDING-03, Medium).

**Location:** `src/auth.rs:706-707`, `src/commands/user.rs:1193-1194`

The first-user-is-admin pattern uses a non-atomic read-then-write:
```rust
let is_first = !has_any_users(&state.pool).await.unwrap_or(true);
let role = if is_first { "admin" } else { "user" };
// INSERT follows separately
```
Two concurrent registrations on a fresh install can both observe zero users and both claim admin, creating an unintended second admin account.

**Suggested fix:** compute the role atomically inside the INSERT:
```sql
INSERT INTO users (..., role, ...) VALUES (...,
  CASE WHEN (SELECT COUNT(*) FROM users) = 0 THEN 'admin' ELSE 'user' END,
  ...)
```
Same change needs to land in `src/commands/user.rs` for parity.

Reported by @marcotama.

olivierlambert:
## Summary
Three code paths assigned the first-user-is-admin role with a TOCTOU pattern: a separate \`has_any_users()\` read before the \`INSERT\`. Two concurrent registrations on a fresh DB could both observe an empty users table and both claim admin.

This PR replaces each read-then-write with a single \`INSERT\` whose role column is computed via a \`CASE\` subquery: \`'admin'\` if no users exist (or, in the CLI, if \`--admin\` was passed), else \`'user'\`. SQLite serializes writers, so the CASE evaluation and the row insert run under the same write lock — the second concurrent registration sees the row count = 1 and gets \`'user'\`.

## Sites fixed
- \`register_handler\` in \`src/auth.rs\` (local registration form)
- \`find_or_create_oidc_user\` in \`src/auth.rs\` (OIDC auto-register)
- \`UserCommands::Create\` in \`src/commands/user.rs\` (CLI \`calrs user create\`)

The CLI uses \`INSERT ... RETURNING role\` so the success message can still distinguish "first user, auto-promoted" from an explicit \`--admin\` invocation.

## Tests added
- \`oidc_first_auto_registered_user_gets_admin_role\`
- \`oidc_subsequent_auto_registered_user_gets_user_role\`
- \`first_user_create_gets_admin_role_atomically\`
- \`subsequent_user_create_gets_user_role\`
- \`explicit_admin_flag_overrides_user_role\`

(The tests pin role-assignment correctness, not concurrency. SQLite write serialization is the language guarantee.)

## Test plan
- [x] \`cargo build\`
- [x] \`cargo test\` (598 passing — was 593, +5 new)
- [ ] Reviewer: run the CLI on a fresh DB, confirm \`calrs user create\` (no \`--admin\`) still shows the auto-promotion message; confirm \`calrs user create --admin\` does not.

Closes #79

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -703,24 +703,26 @@ async fn register_handler(
         Err(_) => return render_register_error(&state, "Internal error", &auth_config),
     };
 
-    // First user gets admin role
-    let is_first = !has_any_users(&state.pool).await.unwrap_or(true);
-    let role = if is_first { "admin" } else { "user" };
-
     let user_id = uuid::Uuid::new_v4().to_string();
     let username = match generate_username(&state.pool, &form.email).await {
         Ok(u) => u,
         Err(_) => return render_register_error(&state, "Internal error", &auth_config),
     };
 
+    // The role is computed inside the INSERT so the "first user gets admin"
+    // logic is atomic with the row creation. A separate read-then-write
+    // would let two concurrent registrations both observe zero users and
+    // both claim admin.
     if sqlx::query(
-        "INSERT INTO users (id, email, name, timezone, password_hash, role, auth_provider, username) VALUES (?, ?, ?, 'UTC', ?, ?, 'local', ?)",
+        "INSERT INTO users (id, email, name, timezone, password_hash, role, auth_provider, username)
+         VALUES (?, ?, ?, 'UTC', ?,
+                 CASE WHEN (SELECT COUNT(*) FROM users) = 0 THEN 'admin' ELSE 'user' END,
+                 'local', ?)",
     )
     .bind(&user_id)
     .bind(&form.email)
     .bind(&form.name)
     .bind(&password_hash)
-    .bind(role)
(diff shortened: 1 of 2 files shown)


olivierlambert:
Pushed follow-ups for the three review suggestions:

1. **`COUNT(*) = 0` → `NOT EXISTS (SELECT 1 FROM users)`** across all three sites. O(1) instead of O(n).
2. **Extracted `auth::create_local_user`** as a `pub(crate)` helper. Both the web registration handler and the CLI `Create` command go through it. The CLI test no longer duplicates SQL; it calls the helper directly, so the test exercises the production code path.
3. **Removed `has_any_users()`** and its two unit tests. No production callers remained after this PR's changes.

Net: 596 tests pass (was 598, -2 removed has_any_users tests). Clippy clean.

PR #90, opened 2026-05-04
fix(web): trust the rightmost X-Forwarded-For for rate limiting

### LINKED ISSUE #81: auth/web: rate limiter trusts attacker-controlled X-Forwarded-For
From the [security audit](https://github.com/user-attachments/files/27326622/SECURITY-REPORT.md) (FINDING-05, Medium).

**Locations** (all use the same pattern, taking the leftmost value):
- `src/auth.rs:529` (login)
- `src/web/mod.rs:7283` (booking)
- `src/web/mod.rs:8059`
- `src/web/mod.rs:8750`
- `src/web/mod.rs:10547`
- `src/web/mod.rs:13048`

```rust
let client_ip = headers
    .get("x-forwarded-for")
    .and_then(|v| v.to_str().ok())
    .and_then(|s| s.split(',').next())  // first = attacker-controlled
    ...
```

`X-Forwarded-For` is appended by each proxy in the chain, so the leftmost value is whatever the original client sent and is fully attacker-controlled. An attacker can rotate `X-Forwarded-For` per request and bypass per-IP rate limits entirely.

**Suggested fix:** take the **rightmost** value (which is set by the most recent trusted hop), or use `X-Real-IP` set by the reverse proxy. Document that the rate limiter is only effective behind a correctly-configured reverse proxy. The `axum-client-ip` crate handles this well (RFC 7239 `Forwarded` parsing + a configurable trusted-proxies list).

Should be factored into a single helper since the pattern is duplicated 6 times.

Reported by @marcotama.

olivierlambert:
## Summary
All 6 rate-limiter sites took the **leftmost** \`X-Forwarded-For\` value, which is exactly the attacker-controlled one: each proxy in the chain appends the IP it observed, so the leftmost is whatever the original request arrived with. An attacker rotating the header per request bypassed per-IP login + booking rate limits entirely.

This PR consolidates the 6 copy-pasted extractions into a single helper and changes the algorithm to:

1. Prefer \`X-Real-IP\` if set (a single value set by the proxy from the TCP peer; the proxy ignores any client-supplied value).
2. Otherwise take the **rightmost** \`X-Forwarded-For\` value (most-recent trusted hop's view of its peer).
3. Fall back to \`"unknown"\`.

Trust model: one reverse proxy in front of calrs, matching the documented Caddy/Nginx deployment.

## Sites consolidated
- \`login_handler\` in \`src/auth.rs\`
- 5 booking-related handlers in \`src/web/mod.rs\`

All now call \`client_ip_for_rate_limit(&headers)\`.

## Tests added
- \`client_ip_unknown_when_no_headers\`
- \`client_ip_prefers_x_real_ip\` (with both headers set, X-Real-IP wins)
- \`client_ip_takes_rightmost_xff\` (the actual bug fix)
- \`client_ip_xff_single_value\`
- \`client_ip_falls_through_empty_x_real_ip\`
- \`client_ip_xff_trims_whitespace\`

## Test plan
- [x] \`cargo build\`
- [x] \`cargo test\` (602 passing — was 596, +6 new)
- [x] Reviewer: behind your reverse proxy of choice, verify login rate limit triggers per-IP rather than per-arbitrary-header value.

Closes #81

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -296,7 +296,7 @@ File: `src/web/mod.rs`, templates in `templates/`
 
 **Security hardening (1.0):**
 - **CSRF protection** — double-submit cookie pattern on all 31 POST handlers via `csrf_cookie_middleware`. Client-side JS injects `_csrf` hidden field. Multipart forms use query parameter.
-- **Booking rate limiting** — per-IP (10 req / 5 min) on all 4 booking handlers. Client IP is sourced from `X-Real-IP` if set, else the *rightmost* `X-Forwarded-For` value (the trusted proxy's view of its peer; leftmost is attacker-controlled). See `client_ip_for_rate_limit()` in `web/mod.rs`.
+- **Booking rate limiting** — per-IP (10 req / 5 min) on all 4 booking handlers. Client IP is the *rightmost* `X-Forwarded-For` value (the trusted proxy's view of its peer; leftmost is attacker-controlled). See `client_ip_for_rate_limit()` in `web/mod.rs`.
 - **Input validation** — server-side on all booking forms (name 1–255, email format, notes max 5000, date max 365 days), registration, settings, avatar upload (content-type whitelist).
 - **Double-booking prevention** — partial unique index `idx_bookings_no_overlap` on `(event_type_id, start_at)` + `BEGIN IMMEDIATE` transactions.
 - **Crash-proof handlers** — all `.unwrap()` in web handlers replaced with proper error responses.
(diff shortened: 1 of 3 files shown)


olivierlambert:
Pushed self-review follow-up:

1. **Dropped `X-Real-IP` support entirely.** As called out in the review, neither default Caddy nor default Nginx overwrites a client-supplied `X-Real-IP`, so trusting it would have let an attacker forge their identity with a single header. Worse than the bug this PR fixes. Now uses rightmost `X-Forwarded-For` only, which is what the documented deployment actually configures.

2. **Replaced the "prefers X-Real-IP" test** with a negative test (`client_ip_ignores_x_real_ip`) that pins the new behavior: even if X-Real-IP is set, we don't trust it.

3. **Removed the em dashes** from the new doc-comment per project preference.

601 tests passing, clippy clean.

PR #92, opened 2026-05-04
fix(admin): reject non-http(s) schemes for company_link

### LINKED ISSUE #83: admin: validate company_link URL scheme to block javascript: URIs
From the [security audit](https://github.com/user-attachments/files/27326622/SECURITY-REPORT.md) (FINDING-07, Medium).

**Location:** `src/web/mod.rs:141-151` (`get_company_link`), and the admin update handler that writes the value.

The admin-configurable `company_link` is stored and rendered as a clickable anchor on every public booking page without scheme validation. An admin (or an attacker who has compromised an admin account) can set it to a `javascript:` URI; the script then runs for every visitor who clicks the link.

**Suggested fix:** allowlist the scheme on both write (admin update handler) and read (defense in depth). Reject non-`http(s)` schemes:
```rust
fn is_safe_url(url: &str) -> bool {
    let lower = url.trim().to_ascii_lowercase();
    lower.starts_with("https://") || lower.starts_with("http://")
}
```
Also worth applying to any other admin-controlled URL fields (review needed).

Reported by @marcotama.

olivierlambert:
## Summary
The admin-controlled \`company_link\` is rendered as a clickable anchor on every public booking page. Without scheme validation, an admin (or an attacker who has compromised an admin account) could set it to \`javascript:alert(1)\` and turn the link into stored XSS for every visitor.

This PR validates the scheme on both ends:

- **Write** (\`admin_update_company_link\`): rejects anything that isn't \`http://\` or \`https://\` and redirects back to \`/dashboard/admin?error=...\` so the admin sees the rejection.
- **Read** (\`get_company_link\`): silently drops a non-http(s) value as defense-in-depth, in case a bad value slipped in via a manual DB edit or older release.

## Why both ends
The write-side check is the primary gate. The read-side check exists because the admin UI was the original ingress and we can't rule out other paths historically writing the column (e.g. a future migration, manual sqlite-cli edit). Cheap belt-and-suspenders.

## Tests added
- \`company_link_accepts_http_and_https\` (valid cases, including upper-case scheme + whitespace)
- \`company_link_rejects_javascript_uri\` (the actual attack)
- \`company_link_rejects_other_dangerous_schemes\` (\`data:\`, \`vbscript:\`, \`file:\`)
- \`company_link_rejects_unschemed_input\` (bare hostname, paths, protocol-relative \`//\`)

## Test plan
- [x] \`cargo build\`
- [x] \`cargo test\` (611 passing)
- [ ] Reviewer: as admin, try saving \`javascript:alert(1)\` in the company-link field. Confirm the admin panel surfaces \"Company link must start with http:// or https://\" and the value is not persisted.

Closes #83

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -152,6 +152,45 @@ dependencies = [
 ]
 
 [[package]]
+name = "asn1-rs"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7f6fd5ddaf0351dff5b8da21b2fb4ff8e08ddd02857f0bf69c47639106c0fff0"
+dependencies = [
+ "asn1-rs-derive",
+ "asn1-rs-impl",
+ "displaydoc",
+ "nom 7.1.3",
+ "num-traits",
+ "rusticata-macros",
+ "thiserror 1.0.69",
+ "time",
+]
+
+[[package]]
+name = "asn1-rs-derive"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "726535892e8eae7e70657b4c8ea93d26b8553afb1ce617caee529ef96d7dee6c"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 1.0.109",
+ "synstructure 0.12.6",
+]
+
+[[package]]
+name = "asn1-rs-impl"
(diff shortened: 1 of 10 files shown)


olivierlambert:
Wrong branch — recreating from fix/company-link-scheme. Sorry for the noise.

PR #94, opened 2026-05-04
fix(auth): encrypt OIDC client_secret at rest

### LINKED ISSUE #84: auth: encrypt OIDC client_secret at rest like CalDAV/SMTP credentials
From the [security audit](https://github.com/user-attachments/files/27326622/SECURITY-REPORT.md) (FINDING-08, Medium).

**Location:** `src/auth.rs:851-854`, `src/commands/config.rs:55`

CalDAV and SMTP passwords are encrypted at rest with AES-256-GCM (`crypto::encrypt_password` / `decrypt_password`). The OIDC `client_secret` in `auth_config` is **not** encrypted; it is stored and read as plaintext:
```rust
let client_secret = auth_config
    .oidc_client_secret
    .as_ref()
    .map(|s| ClientSecret::new(s.clone()));  // plaintext from DB
```
A SQLite file leak directly exposes the OIDC client secret, which lets the attacker impersonate calrs to the IdP.

**Suggested fix:**
1. Reuse `crypto::encrypt_password` / `decrypt_password` for `oidc_client_secret` on write/read paths in `src/auth.rs` and `src/commands/config.rs`.
2. Add a one-shot startup migration mirroring `db::migrate_passwords()` to re-encrypt any existing plaintext value (detect by absence of the GCM ciphertext envelope).

Reported by @marcotama.

olivierlambert:
## Summary
CalDAV and SMTP passwords are encrypted with AES-256-GCM, but the OIDC `client_secret` in `auth_config` was stored as plaintext. Any SQLite file leak directly exposed it, letting an attacker impersonate calrs to the IdP.

This PR closes the gap.

## What changes

- **New crypto API** (`src/crypto.rs`):
  - `encrypt_value` / `decrypt_value` / `is_encrypted_value` with a sentinel `enc:v1:` prefix so encrypted values are unambiguously distinguishable from plaintext at migration time.
  - Why a separate API instead of reusing `encrypt_password`: an OIDC plaintext secret can coincidentally look like base64 (the format `encrypt_password` produces). The prefix removes the ambiguity.

- **Read site** (`src/auth.rs::build_oidc_client_with_redirect`):
  - Now takes `secret_key: &[u8; 32]` and decrypts the stored value before wrapping it in `ClientSecret`. Both callers (`oidc_login`, `oidc_callback`) already have `state.secret_key` in scope.

- **Write sites** (3): all encrypt before binding:
  - `src/web/mod.rs::admin_update_oidc` (admin panel)
  - `src/commands/config.rs` interactive `Oidc` (CLI full setup)
  - `src/commands/config.rs --client-secret` (CLI single-field update)

- **Startup migration** (`src/db.rs::migrate_passwords`): mirrors the existing CalDAV/SMTP password migration. If an `auth_config.oidc_client_secret` value is present and doesn't carry the `enc:v1:` prefix, it gets encrypted in place. Idempotent: the prefix check makes a second run a no-op.

## Tests added
- `crypto::test_encrypt_value_roundtrip`
- `crypto::test_is_encrypted_value` (specifically asserts that base64-looking plaintext is NOT identified as encrypted)
- `crypto::test_decrypt_value_rejects_unprefixed`
- `db::migrate_passwords_encrypts_plaintext_oidc_secret`
- `db::migrate_passwords_is_idempotent_for_oidc_secret`

## Test plan
- [x] `cargo build`
- [x] `cargo test` (616 passing — was 611, +5 new)
- [ ] Reviewer: on a DB with an existing plaintext OIDC secret, restart calrs and confirm the value gets the `enc:v1:` prefix while OIDC login still works.

Closes #84

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -836,6 +836,7 @@ fn build_http_client() -> Result<openidconnect::reqwest::Client> {
 
 async fn build_oidc_client_with_redirect(
     auth_config: &AuthConfig,
+    secret_key: &[u8; 32],
 ) -> Result<
     CoreClient<
         EndpointSet,
@@ -863,10 +864,19 @@ async fn build_oidc_client_with_redirect(
             .clone(),
     );
 
+    // The stored OIDC client secret is encrypted at rest. The startup
+    // password migration (db::migrate_passwords) handles converting
+    // pre-existing plaintext values, so by the time we reach here the
+    // value should always carry the enc:v1: prefix. Skip silently if not
+    // (a malformed value would just produce a request without a secret).
     let client_secret = auth_config
         .oidc_client_secret
         .as_ref()
-        .map(|s| ClientSecret::new(s.clone()));
+        .filter(|s| !s.is_empty())
+        .map(|s| crate::crypto::decrypt_value(secret_key, s))
+        .transpose()
+        .map_err(|e| anyhow::anyhow!("OIDC client secret decryption failed: {}", e))?
+        .map(ClientSecret::new);
 
     let http_client = build_http_client()?;
     let provider_metadata = CoreProviderMetadata::discover_async(issuer_url, &http_client)
@@ -891,7 +901,7 @@ async fn oidc_login(State(state): State<Arc<AppState>>) -> Response {
         _ => return Html("OIDC is not enabled.".to_string()).into_response(),
     };
 
-    let client = match build_oidc_client_with_redirect(&auth_config).await {
+    let client = match build_oidc_client_with_redirect(&auth_config, &state.secret_key).await {
         Ok(c) => c,
         Err(e) => return crate::web::oidc_error_response("oidc client build (login)", &e),
     };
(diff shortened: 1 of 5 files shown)


olivierlambert:
Self-review follow-up pushed:

1. **Comment fix** in `build_oidc_client_with_redirect`: the old version implied a fallback-to-no-secret behavior that does not exist. New comment matches what the code actually does (surface a decryption failure as an OIDC error so the admin sees a clear failure rather than silently broken sign-ins).

2. **Try-decrypt belt-and-suspenders** in the migration: the "is encrypted?" check is no longer purely structural. It now also tries to decrypt; the value is skipped only if the prefix matches AND decryption succeeds under the current key. A plaintext that coincidentally starts with `enc:v1:` now gets correctly re-encrypted instead of being mistaken for already-migrated. Ciphertext from a rotated key was already unrecoverable, so re-encrypting doesn't lose anything new.

Added `migrate_passwords_handles_plaintext_starting_with_prefix` covering the pathological prefix-collision case.

617 tests passing (was 616, +1 new). Clippy clean.

PR #96, opened 2026-05-04
feat: minimum notice for guest cancel and reschedule

### LINKED ISSUE #95: Add minimum notice period for guest-initiated cancel and reschedule
Originally requested in #88.

## Background

Currently, guests can cancel or reschedule a confirmed booking via the tokenized links in their confirmation email at any time, including a few minutes before the meeting starts. This leaves hosts with same-day calendar holes that cannot be re-filled.

Neither Calendly nor Cal.com solves this today. Calendly only displays a free-text policy without enforcement; Cal.com offers a binary "disable cancel / disable reschedule" toggle without a time window (open feature request: [calcom/cal.com#15647](https://github.com/calcom/cal.com/issues/15647), community PR closed unmerged). A per-event-type minimum-notice setting is a small but meaningful differentiator for calrs.

## Proposed behavior

Add two optional per-event-type settings:

- `cancel_notice_min` minimum minutes between "now" and the booking start required for a **guest** to self-cancel.
- `reschedule_notice_min` same, for guest-initiated reschedule.

Both default to `NULL` (no restriction, current behavior). Set per event type from the event type form.

Restrictions apply to **guest-side actions only** (the token endpoints):

- `/booking/cancel/{cancel_token}` (GET form + POST)
- `/booking/reschedule/{reschedule_token}` (GET + POST)

Hosts and admins continue to cancel/reschedule freely from the dashboard. Hosts often need to act on real-world emergencies on behalf of a guest, so the lead-time check does not apply there.

## Schema change

```sql
-- migrations/0NN_cancel_reschedule_notice.sql
ALTER TABLE event_types ADD COLUMN cancel_notice_min INTEGER;
ALTER TABLE event_types ADD COLUMN reschedule_notice_min INTEGER;
```

Both nullable. `NULL` or `0` means no restriction. Register the migration in `src/db.rs` per the convention in `CLAUDE.md` (forgetting this step has caused production bugs before).

## Validation

In the guest cancel and reschedule handlers in `src/web/mod.rs`:

```rust
if let Some(notice_min) = event_type.cancel_notice_min {
    if notice_min > 0 {
        let cutoff = booking.start_at - chrono::Duration::minutes(notice_min as i64);
        if chrono::Utc::now() > cutoff {
            return render_blocked_page(notice_min, &host_email);
        }
    }
}
```

Same shape for reschedule. Both checks need to live before any state mutation.

## UI

1. **Event type form** (`templates/event_type_form.html`): two new fields next to the existing notice/limit settings ("Minimum notice to cancel", "Minimum notice to reschedule") with a unit selector (minutes/hours/days). Empty = no restriction.
2. **New template** `templates/booking_action_blocked.html`: friendly page with the host's email as a `mailto:` link. Something like: *"This booking can no longer be cancelled online. The host requires X hours of notice. Please contact `host@example.com` directly if you cannot attend."*
3. **Booking confirmed page** and **confirmation email**: when a notice period is set, surface it inline near the cancel/reschedule buttons so guests know upfront. Avoids the surprise of clicking a button and being told "no".

## i18n

Land new strings on the `i18n` branch first per the workflow in `CLAUDE.md`. New keys roughly:

- `event-type-form.cancel-notice-label`
- `event-type-form.reschedule-notice-label`
- `event-type-form.notice-help` ("Leave empty for no restriction")
- `booking-blocked.title`
- `booking-blocked.body` (with host email and notice window args)

## Out of scope for this issue

- Per-team or per-host override
- Different lead times by booking tier or value
- Payment-as-cancellation-deterrent (Stripe integration)
- Reason-based gating ("only allow cancel if reason is provided")

These can be revisited later if demand emerges.

## Acceptance criteria

- [x] Migration added and registered in `src/db.rs`
- [x] `cancel_notice_min` and `reschedule_notice_min` settable from the event type form
- [x] Guest cancel via tokenized link is blocked inside the notice window, with a friendly page showing the host's contact
- [x] Guest reschedule via tokenized link is blocked under the same rule
- [x] Host and admin cancel/reschedule from the dashboard unaffected
- [x] Notice policy surfaced on confirmation email and booking-confirmed page when set
- [x] Strings added to `i18n/en/main.ftl` on the `i18n` branch
- [x] Tests cover: no-notice case (works), within-window (blocked), outside-window (works), host bypass

olivierlambert:
Closes #95.

## Summary

Hosts can now require a minimum lead time before a guest can self-cancel or self-reschedule via the tokenized email links, set per event type from the event type form. `NULL` or `0` means no restriction (current behavior).

Within the notice window, the four guest token endpoints render a friendly `booking_action_blocked` page with the host's contact email instead of mutating booking state. Host- and admin-initiated cancellations from the dashboard are unaffected.

The policy is also surfaced inline on the confirmed page and in the confirmation email so guests aren't surprised at click time.

## What changed

| Area | File(s) | Change |
|---|---|---|
| Schema | `migrations/050_min_notice_cancel_reschedule.sql` | New `cancel_notice_min` + `reschedule_notice_min` INTEGER columns on `event_types` |
| Schema | `src/db.rs` | Migration registered (per CLAUDE.md) |
| Model | `src/models.rs` | `EventType.cancel_notice_min: Option<i32>` + `reschedule_notice_min: Option<i32>` |
| Validation | `src/web/mod.rs` | New `check_notice_window` helper called before any state mutation in `guest_cancel_form` (GET), `guest_cancel_booking` (POST), `guest_reschedule_slots` (GET), `guest_reschedule_booking` (POST) |
| Form | `src/web/mod.rs` + `templates/event_type_form.html` | Numeric input + minutes/hours/days select, value/unit converted on submit, natural unit recovered on edit |
| Confirmed page | `src/web/mod.rs` + `templates/confirmed.html` | Policy surfaced inline when set |
| Email | `src/email.rs` | Notice line appended to plain text + HTML when set |
| Blocked page | `templates/booking_action_blocked.html` (new) | Friendly page with host's `mailto:` link |
| i18n | `i18n/en/main.ftl` | 11 new English keys (other locales fall back to English at runtime) |

The hardest piece is the contract between handler context and templates: the form submits `(cancel_notice_value, cancel_notice_unit)` and `(reschedule_notice_value, reschedule_notice_unit)`; backend converts to minutes and stores. On render, backend computes the natural unit (1440-divisible → days, 60-divisible → hours, else minutes) and passes that back to the form for display.

## Worth a careful look

- **Time semantics**: `bookings.start_at` is stored as a naive host-local string, not UTC. The check uses `Local::now().naive_local()` rather than `Utc::now()`, mirroring the existing `slot_start < now + Duration::minutes(min_notice as i64)` precedent in `guest_reschedule_booking`. The issue's example used `Utc::now()` but the file's own conventions are host-local.
- **`booking_action_blocked.html` uses `| safe`** on the `booking-blocked-body` Fluent key because the body contains an inline `<a href="mailto:...">`. The `host_email` placeholder is sourced from the user record (validated at registration), so practical XSS surface is small, but a stricter version could split the template (escape the email separately, keep the `<a>` from Fluent). Flagged as a follow-up rather than a blocker.

## Tests

7 new tests in `src/web/mod.rs::tests`:

- `guest_cancel_no_notice_succeeds`
- `guest_cancel_within_notice_window_blocked` (covers GET + POST)
- `guest_cancel_outside_notice_window_succeeds`
- `guest_reschedule_no_notice_succeeds`
- `guest_reschedule_within_notice_window_blocked` (covers GET + POST)
- `guest_reschedule_outside_notice_window_succeeds`
- `host_cancel_unaffected_by_notice` (host route bypasses the check)

Test count went from 617 → 624. Helper `insert_booking_at(pool, et_id, cancel_tok, resched_tok, offset)` writes a booking at `Local::now() + offset` so the tests exercise real time semantics rather than fixed dates.

## Test plan

- [x] `cargo build`
- [x] `cargo fmt --check`
- [x] `cargo test` (624 passing)
- [ ] Reviewer: create an event type with a 60-minute cancel notice; book it 30 min in the future; click the cancel link in the email and confirm you land on the blocked page with the correct host email. Then book it 2 hours out and confirm cancel still works.
- [ ] Reviewer: confirm the host's own dashboard cancel of a booking inside the guest notice window still works (the check is guest-side only).

## Acceptance criteria from #95

- [x] Migration added and registered in `src/db.rs`
- [x] `cancel_notice_min` and `reschedule_notice_min` settable from the event type form
- [x] Guest cancel via tokenized link blocked inside the notice window with a friendly host-contact page
- [x] Guest reschedule via tokenized link blocked under the same rule
- [x] Host and admin cancel/reschedule from the dashboard unaffected
- [x] Notice policy surfaced on confirmation email and booking-confirmed page when set
- [x] Strings added to `i18n/en/main.ftl` (this PR's `main` already includes the recent `i18n → main` merge)
- [x] Tests cover all four cases listed in the issue

### DIFF
--- a/i18n/en/main.ftl
+++ b/i18n/en/main.ftl
@@ -223,6 +223,11 @@ email-cancel-headline-by-guest = Your booking has been cancelled.
 email-cancel-ics-attached-plain = A calendar cancellation is attached.
 email-cancel-ics-attached-html = A calendar cancellation is attached to this email.
 
+# Confirmation email: notice-window policy lines (src/email.rs)
+
+email-confirm-cancel-notice = Note: cancellation requires at least { $minutes } minutes notice.
+email-confirm-reschedule-notice = Note: rescheduling requires at least { $minutes } minutes notice.
+
 # Event type form: cancel/reschedule minimum notice (templates/event_type_form.html)
 
 event-type-form-cancel-notice-label = Minimum notice to cancel
--- a/i18n/en/main.ftl
+++ b/i18n/en/main.ftl
@@ -222,3 +222,23 @@ email-cancel-headline-by-host = Your booking has been cancelled by { $host }.
 email-cancel-headline-by-guest = Your booking has been cancelled.
 email-cancel-ics-attached-plain = A calendar cancellation is attached.
 email-cancel-ics-attached-html = A calendar cancellation is attached to this email.
+
+# Event type form: cancel/reschedule minimum notice (templates/event_type_form.html)
+
+event-type-form-cancel-notice-label = Minimum notice to cancel
+event-type-form-reschedule-notice-label = Minimum notice to reschedule
+event-type-form-notice-help = Leave empty for no restriction.
+event-type-form-notice-unit-minutes = minutes
+event-type-form-notice-unit-hours = hours
+event-type-form-notice-unit-days = days
+
+# Booking confirmation: cancel/reschedule policy notices (templates/confirmed.html)
+
+confirmed-cancel-notice-info = Cancellation requires at least { $minutes } minutes notice before the meeting.
+confirmed-reschedule-notice-info = Rescheduling requires at least { $minutes } minutes notice before the meeting.
+
+# Booking action blocked page (templates/booking_action_blocked.html)
+
+booking-blocked-title-cancel = This booking can no longer be cancelled online
+booking-blocked-title-reschedule = This booking can no longer be rescheduled online
+booking-blocked-body = The host requires at least { $minutes } minutes of notice. If you cannot attend, please email <a href="mailto:{ $host_email }">{ $host_email }</a> directly.
--- a/migrations/050_min_notice_cancel_reschedule.sql
+++ b/migrations/050_min_notice_cancel_reschedule.sql
@@ -1,0 +1,6 @@
+-- Per-event-type minimum notice window for guest-initiated cancel and reschedule.
+-- NULL or 0 means no restriction (the existing behaviour).
+-- Values are stored in minutes; the form lets the user pick minutes/hours/days
+-- and converts to minutes before INSERT/UPDATE.
+ALTER TABLE event_types ADD COLUMN cancel_notice_min INTEGER;
+ALTER TABLE event_types ADD COLUMN reschedule_notice_min INTEGER;
--- a/src/db.rs
+++ b/src/db.rs
@@ -215,6 +215,10 @@ pub async fn migrate(pool: &SqlitePool) -> Result<()> {
             "049_smtp_decouple_account",
             include_str!("../migrations/049_smtp_decouple_account.sql"),
         ),
+        (
+            "050_min_notice_cancel_reschedule",
+            include_str!("../migrations/050_min_notice_cancel_reschedule.sql"),
+        ),
     ];
 
     let mut applied_count = 0u32;
@@ -796,7 +800,7 @@ mod tests {
             .fetch_one(&pool)
             .await
             .unwrap();
-        assert_eq!(count.0, 49, "All 49 migrations should be tracked");
+        assert_eq!(count.0, 50, "All 50 migrations should be tracked");
     }
 
     #[tokio::test]
@@ -810,7 +814,7 @@ mod tests {
             .fetch_one(&pool)
             .await
             .unwrap();
-        assert_eq!(count.0, 49, "Still 49 migrations after second run");
+        assert_eq!(count.0, 50, "Still 50 migrations after second run");
     }
 
     #[tokio::test]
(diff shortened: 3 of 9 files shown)


olivierlambert:
Pushed self-review follow-up:

**Email notice lines now localized.** Added `email-confirm-cancel-notice` and `email-confirm-reschedule-notice` Fluent keys (parameterized on `$minutes`) and switched the two `format!` calls in `src/email.rs` to `ta(lang, ...)`. A non-English guest now gets the notice line in their locale instead of the only English line in an otherwise translated email.

Other follow-ups from the self-review (combined notice line UX, email body content test, blocked-page `| safe` belt-and-suspenders) are deferred since they're cosmetic / defense-in-depth, not blockers.

PR #104, opened 2026-05-14
Fix mailbox and refactorize

aprilmay:
This PR fixes a generic error sending emails, eg.:

```sh
calrs config smtp-test calrs@example.com
… Sending test email to calrs@example.com…
Error: Invalid input
```

The handling of the `from` mailbox is refactorized in all functions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -138,6 +138,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/), and this
 | Estonian locale | 1.10.0 | First community-language slot added beyond the original four (English, French, Spanish, Polish, German, Italian) |
 | Admin user deletion | 1.10.0 | Admins can permanently delete users from the admin panel with cascade rules and confirmation |
 
+## [1.10.2] - 2026-05-14
+
+Hotfix for a production incident in which CalDAV sync was wrongly cancelling live customer bookings. No schema changes, no behaviour changes outside the orphan-cancellation surface.
+
+### Fixed
+
+- **Stop cancelling bookings on phantom sync-collection "deletions"** — `delete_events_by_href` (src/commands/sync.rs:443) cancelled any booking whose UID matched an href the server reported as deleted, regardless of whether the local `events` table had a matching row. In production, BlueMind's sync-collection emitted a "deleted" entry for an href whose local event lived on a different calendar (the booking's write calendar); `cancel_orphaned_booking`'s global UID-only lookup still found the confirmed booking and cancelled it + emailed host and guest "the event was deleted by the host." The cancellation is now gated on `rows_affected > 0` — if we never had a local event for that calendar/href, that's a server-side false positive and we log a warning instead. Two regression tests added: one captures the exact prod failure mode, the other guards the legitimate deletion case from regressing
+
+### Internal
+
+- 634 tests total (up from 632 in 1.10.1), all green on pre-commit
+- Three follow-ups tracked separately for the next release: confirm-before-cancel via HEAD on the resource href (#105), source/account scoping in `cancel_orphaned_booking` (#106), propstat-aware sync-collection XML parser (#107)
+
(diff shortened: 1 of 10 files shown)


olivierlambert:
Thanks for catching this! The root cause analysis is right: when `from_name` is `None`, the old fallback put `from_email` as display name, and the `@` made the whole mailbox string fail RFC 5322 parsing. Switching to `Mailbox::new(Option<String>, Address)` is the correct fix and also makes us more robust when `from_name` contains chars that need quoting (commas, periods, non-ASCII).

A few small follow-ups before merge:

1. **Use the helper consistently in `send_test_email`.** It currently builds the mailbox inline:
   ```rust
   let from = Mailbox::new(config.from_name.to_owned(), config.from_email.parse()?);
   ```
   Since that's exactly what `from_mailbox()` does, please swap it for `config.from_mailbox()?` — the whole point of introducing the helper is to have one call site shape everywhere.

2. **Drop (or generalize) the leftover debug line.** `tracing::debug!("Sending: {:?}", email)` only exists in `send_test_email`. It looks like an investigation artifact. Either remove it, or move an equivalent log into `send_email()` so all sends are traced uniformly.

3. **Add a small unit test for `SmtpConfig::from_mailbox()`.** Two cases would lock this in cheaply:
   - `from_name = None` (the exact case that produced `Error: Invalid input`)
   - `from_name = Some("Name, With Comma")` to confirm proper quoting

4. **Functional sanity check.** Could you also do a manual end-to-end run with both configurations and confirm the `From:` header looks right in the received message? Roughly:
   - `from_name = None` → `calrs config smtp-test you@example.com` should succeed and the header should be just `From: calrs@example.com` (no bracketed display name).
   - `from_name = Some("Some Name")` → header should be `From: Some Name <calrs@example.com>` in the received mail.
   That confirms we didn't regress the happy path while fixing the broken one.

Nits (non-blocking):
- `to_owned()` on `Mailbox` / `Option<String>` is fine but `clone()` is the more idiomatic spelling since those types are `Clone`.
- The doc comment on `from_mailbox()` could mention the intent: building the From mailbox from config, omitting the display name when `from_name` is unset.

Otherwise this is a clean, focused fix — nice cleanup of a pattern that was duplicated 17 times. Thanks!

aprilmay:
Updated with elements of the reply, basic manual functional sanity check passed

olivierlambert (approved):
Thanks @aprilmay, this is in great shape now! 🙏 Going through the items from my earlier comment:

- ✅ `send_test_email` uses `config.mailbox_from()?` (no more inline `format!`)
- ✅ Unit test `smtp_config_mailbox_from` covers both the `from_name = None` case (the one that originally produced `Error: Invalid input`) and the `Name, With Comma` quoting case
- ✅ Manual functional sanity check confirmed per your reply
- ✅ `clone()` used in the helper

On the helper name: I had originally suggested `from_mailbox()`, but on reflection `mailbox_from()` is actually the better choice. `from_*` reads as a `From`-trait-style constructor in Rust ("build a SmtpConfig from a mailbox"), which is the opposite of what this does. `mailbox_from()` correctly parses as "the mailbox for the From header". Nice call.

On the leftover `tracing::debug!("Sending: {:?}", email)` in `send_test_email`: I see you scoped it with a comment explaining why. Fine to keep it as-is for this PR. As a small follow-up sometime, I'd lean toward consolidating into `send_email()` so every send is uniformly traceable during local debugging (and we don't print recipient details for production sends). No need to fold into this PR.

Just one tiny tweak before merge, the kind of thing a single character commit fixes:

- `src/email.rs:46` typo in the doc comment: `complient` → `compliant`.

Land that and the PR is good to merge.

`cargo test` passes (651/651), `cargo fmt --check` clean, `cargo clippy` clean. Approving so this can go in as soon as the typo is fixed. Thanks for the thoughtful cleanup, especially deduplicating that pattern across 17 sites 🙂

PR #112, opened 2026-05-15
feat: add 'config dump' CLI command

mvalois:
Add a `calrs config dump [--pretty]` subcommand that prints the configuration in JSON format, from the database: 
```bash
$ calrs config dump --data-dir . --pretty
{
  "auth": {
    "allowed_email_domains": null,
    "created_at": "2026-04-29 14:52:25",
    "oidc_auto_register": true,
    "oidc_client_id": "calrs",
    "oidc_enabled": false,
    "oidc_issuer_url": "https://keycloak.example.com/realms/your-realm",
    "registration_enabled": true,
    "updated_at": "2026-05-15 10:26:09"
  },
  "caldav_sources": [],
  "event_types": [],
  "groups": [],
  "smtp": null,
  "teams": [],
  "users": [
    {
      "allow_dynamic_group": true,
      "auth_provider": "local",
      "bio": null,
      "booking_email": null,
      "created_at": "2026-04-29 19:12:39",
      "email": "mvalois@example.com",
      "enabled": true,
      "id": "513a192d-4b6a-4afe-8164-9f2f4a01d2c0",
      "language": null,
      "name": "MV",
      "oidc_subject": null,
      "role": "admin",
      "timezone": "Europe/Paris",
      "title": null,
      "username": "mvalois"
    }
  ]
}

```

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -305,6 +305,9 @@ pub async fn delete_user(
 
 // --- Auth config helpers ---
 
+// Uses SELECT * intentionally: oidc_client_secret is encrypted at rest (AES-256-GCM)
+// and encrypted in memory. Accidental serialization is prevented by #[serde(skip)]
+// on the AuthConfig.oidc_client_secret field.
 pub async fn get_auth_config(pool: &SqlitePool) -> Result<AuthConfig> {
     let config =
         sqlx::query_as::<_, AuthConfig>("SELECT * FROM auth_config WHERE id = 'singleton'")
--- a/src/commands/config.rs
+++ b/src/commands/config.rs
@@ -1,4 +1,4 @@
-use anyhow::Result;
+use anyhow::{Context, Result};
 use clap::Subcommand;
 use colored::Colorize;
 use sqlx::SqlitePool;
@@ -42,6 +42,12 @@ pub enum ConfigCommands {
         #[arg(long)]
         allowed_domains: Option<String>,
     },
+    /// Dump all configuration as JSON
+    Dump {
+        /// Pretty-print the JSON output
+        #[arg(long)]
+        pretty: bool,
+    },
     /// Configure OIDC (OpenID Connect) for SSO login
     Oidc {
         /// OIDC issuer URL (e.g. https://keycloak.example.com/realms/myrealm)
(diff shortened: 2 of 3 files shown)


olivierlambert (commented):
Thanks for this @mvalois! 🙏 First look from me. Really like the security posture: explicit per-table dump structs that only `SELECT` non-secret columns, the defense-in-depth `#[serde(skip)]` added to `AuthConfig.oidc_client_secret` (good belt-and-braces for any future code path that might serialize the whole struct), and tests that explicitly assert `!sources[0].contains_key("password_enc")` and `!user.contains_key("password_hash")`. Treating "no secret in the output" as a tested invariant is exactly the discipline I'd want for a command like this. Good idea splitting `build_dump_output` out as an `async fn -> serde_json::Value` so it's testable without the println side effect, too.

`cargo test` passes (16/16 in the config module), `cargo clippy` clean.

A handful of discussion items below, none of them blocking in isolation, but the first one is worth aligning on before merge.

## 1. Intended scope of "dump all configuration"

The clap help text says `Dump all configuration as JSON` and the dump currently covers `auth`, `smtp`, `users`, `caldav_sources`, `event_types`, `teams`, `groups`. That's a reasonable slice of the top-level entities, but quite a few relational/sub-config tables aren't included:

- `availability_rules` and `availability_overrides` (weekly + per-date schedules per event type, core scheduling config)
- `user_availability_rules` (per-user default working hours)
- `event_type_calendars` (which calendars each event type checks)
- `event_type_member_weights`, `event_type_watchers` (team-event-type config)
- `team_members`, `team_groups` (team composition and OIDC group linkage)
- `user_groups` with weights (OIDC group membership + round-robin priority)
- `booking_frequency_limits`
- `accounts`

Could you say a bit about the intended use case? If it's "give me a glance at the high-level config", the current set is fine and I'd soften the help text from "all configuration" to "configuration summary". If the goal is a building block for backup / audit / restore workflows, the dump probably wants to be more exhaustive (and at that point a top-level `"schema_version"` field would be cheap insurance, see point 4 below).

I'd lean toward the more-exhaustive direction since `config dump` tends to be most valuable as the foundation for `config import` and operator-side backups, but I'm happy to defer that to a follow-up if you'd rather keep this PR tight.

## 2. Inconsistent `enabled` filtering

`users` is filtered to `WHERE enabled = 1` (with a test asserting disabled users don't appear), but `caldav_sources`, `event_types`, and `teams` include disabled rows. For a dump, I'd suggest including everything: a disabled user is still a configured user, and an operator inspecting the config probably wants to see "this user is disabled" rather than "this user doesn't exist". Picking one rule and applying it consistently would make the output less surprising.

## 3. Sync state in `caldav_sources`

`last_synced` and `sync_token` are operational state, not config. They're harmless in an audit context, but if this dump ever feeds a future `config import` on a fresh deployment, transplanting those columns would make calrs lie about its sync state. Worth either dropping them from `CaldavSourceDump` or leaving a comment flagging the distinction.

## 4. Schema version field

A top-level `"schema_version": 1` (or `"calrs_version"`) is cheap insurance against future changes. Easy to add now, painful to add later once tooling depends on the current shape.

## 5. Tiny mechanical thing

`cargo fmt --check` flags the `let smtp_json: serde_json::Value = ...` block in `build_dump_output`. rustfmt wants the match flatter:

```rust
let smtp_json: serde_json::Value = match sqlx::query_as::<_, SmtpConfigDump>(
    "SELECT id, host, port, username, from_email, from_name, enabled \
     FROM smtp_config LIMIT 1",
)
.fetch_optional(pool)
.await?
{
    Some(row) => serde_json::to_value(&row)?,
    None => serde_json::Value::Null,
};
```

The repo's pre-commit hook runs `cargo fmt --check` and will reject the current shape.

Let me know your thinking on #1 in particular and we can go from there. Nice work, and thanks for picking this up 🙂

olivierlambert (changes requested):
Flipping to Changes Requested per the review above. Two actionable items:

1. `cargo fmt --check` fails on the SMTP match block in `build_dump_output`. The pre-commit hook will reject the current shape; the canonical formatting is in the review.
2. Alignment on scope question (1) for "dump all configuration" so we can decide whether to expand coverage or soften the help text.

Nothing controversial, just want to make sure these don't get lost. Thanks!

olivierlambert:
Merged, thank you @mvalois! 🎉

Since the review went unanswered for a couple of weeks and we wanted this in for the next release, I landed your commit as-is and pushed a follow-up commit (898810a) on top addressing the review items:

- went with the exhaustive direction for scope question 1: the dump now covers 18 sections including `availability_rules`, `availability_overrides`, `user_availability_rules`, `event_type_calendars`, `event_type_member_weights`, `event_type_watchers`, `team_members`, `team_groups`, `user_groups`, `booking_frequency_limits` and `accounts`, plus the newer config columns (captcha, meeting links, OAuth2 CalDAV, EWS provider type, SMTP TLS mode)
- disabled users are now included (with their `enabled` flag) for consistency
- operational sync state dropped from `caldav_sources`
- top-level `schema_version: 1` added
- rustfmt fix on the SMTP match block

Your testing discipline around "no secret in the output" is kept and extended: the new sections exclude `access_token_enc`/`refresh_token_enc`, `captcha_secret`, `google_oauth2_client_secret`, `meeting_webhook_secret` and `teams.invite_token`, with tests asserting it.

Thanks again for the contribution, this is a great foundation for a future `config import`.

mvalois:
Well, sorry for not having followed this PR recently, I had work on the workbench...
You've done a great work, thank you again for this merge!
Cheers,
M.

PR #125, opened 2026-06-04
Embed code generator (Inline, Float button, element Click)

arthur-perrot:
This PR adds a Cal.com-style embed system that lets users embed their booking pages on any external website, without any backend changes or API keys.

A single self-contained script (assets/embed.js) that exposes window.Calrs with three methods:

- Calrs.inline({ selector, link, config }) — injects an auto-sizing iframe into an existing container element. Listens to calrs:resize postMessages from the iframe to keep height in sync with content.
- Calrs.floatingButton({ link, buttonText, buttonPosition, buttonColor, textColor, showIcon, config }) — renders a floating pill button (4 corner positions) that opens the booking page in a centered modal overlay.
- Calrs.elementClick() — auto-binds to any element with data-calrs-link / data-calrs-config attributes on DOMContentLoaded. Useful when the host page already has its own CTA buttons or links.

?embed=1 query param — embed mode for booking pages
When embed=1 is present, the booking flow (slots → book → confirmed):
- Strips navigation chrome (logo, footer, theme toggle)
- Sets permissive framing headers (X-Frame-Options: ALLOWALL, frame-ancestors * in CSP)
- Accepts layout, theme, and brand URL params to customize appearance per-embed
- Posts calrs:resize messages so the parent can autosize the iframe
- Uses SameSite=None; Secure for the CSRF cookie so it works correctly in cross-origin iframes (standard Lax is blocked by browsers in third-party frames)
- Propagates embed params through the full booking flow including form submissions and redirects

GET /dashboard/event-types/{slug}/embed — snippet generator
An authenticated dashboard page where users can:

- Pick variant (inline / floating button / element-click) via tabs
- Configure layout, theme, brand color
- Configure floating button appearance (text, position, background/text color, calendar icon)
- Copy the generated <script> snippet
- Preview the embed live in-page (inline as autosizing iframe, floating as scoped button + click-to-open simulation, element-click as clickable demo button + text link)

The Embed button is surfaced directly next to Edit/Disable in the event types list (not buried in the meatballs menu).

### Security

- CSP: embed mode is detected by csp_middleware and swaps frame-ancestors 'self' → frame-ancestors * without breaking captcha allowances
- CSRF: cross-origin iframe detection adds SameSite=None; Secure to the CSRF cookie only in embed context
- Brand color injection: validated as a strict 3 or 6 hex digit string before being interpolated into CSS — no injection vector
- postMessage origin check: the resize listener in embed.js checks ev.source === iframe.contentWindow before acting

### Bug fixes included

- Iframe autosize feedback loop: hovering time slots triggered transitionend events which called send(), which read documentElement.scrollHeight — coupled to the iframe viewport height the parent had set, causing +8px growth per hover. Fixed by measuring document.body only (which is min-height: 0 in embed mode and decoupled from the viewport).

### Important change
Increased embed page container max-width from 900px to 1400px for a wider, more realistic preview experience.
only in embed.html, not all the app

Before, with default values everywhere on the calrs app :
image

after for desktop views : 

image

### DIFF
--- a/assets/embed.js
+++ b/assets/embed.js
@@ -1,0 +1,184 @@
+/* calrs embed runtime — pasted into a host site and exposes window.Calrs.
+ *
+ * API:
+ *   Calrs.inline({ selector, link, config })
+ *   Calrs.floatingButton({ link, buttonText, buttonPosition, buttonColor, textColor, showIcon, config })
+ *   Calrs.elementClick()
+ *
+ * Where `link` is a full URL to a calrs booking page (e.g.
+ * "https://cal.example.com/u/alice/intro") and `config` is an object with
+ * optional `layout` ("month" | "week" | "column"), `theme` ("auto" | "light"
+ * | "dark"), and `brand` (hex color, with or without leading "#"). These end
+ * up as query parameters on the iframe src and feed into the embed-mode
+ * rendering on the calrs side.
+ */
+(function () {
+  if (window.Calrs) return;
+
+  function buildSrc(link, config) {
+    if (!link) return 'about:blank';
+    var url = link + (link.indexOf('?') >= 0 ? '&' : '?') + 'embed=1';
+    if (config) {
+      if (config.layout) url += '&layout=' + encodeURIComponent(config.layout);
+      if (config.theme) url += '&theme=' + encodeURIComponent(config.theme);
+      if (config.brand) {
+        var b = String(config.brand).replace(/^#/, '');
+        url += '&brand=' + encodeURIComponent(b);
+      }
+    }
+    return url;
+  }
+
+  function originOf(url) {
+    try { return new URL(url, location.href).origin; } catch (e) { return null; }
+  }
+
+  // Auto-size: parent listens for {type:'calrs:resize', height} from the
+  // booking page inside the iframe. The booking page emits this on load and
+  // whenever its content height changes. We accept messages only from the
+  // iframe's own origin and contentWindow to avoid trusting unrelated frames.
+  function attachResize(iframe) {
+    var iframeOrigin = originOf(iframe.src);
+    window.addEventListener('message', function (ev) {
+      if (!ev.data || ev.data.type !== 'calrs:resize') return;
+      if (iframeOrigin && ev.origin !== iframeOrigin) return;
+      if (ev.source !== iframe.contentWindow) return;
+      var h = Math.max(60, Math.floor(ev.data.height || 0)) + 8;
+      iframe.style.height = h + 'px';
+    });
+  }
+
+  // Single shared modal — reused by floatingButton and elementClick.
+  var modal = null;
+  function ensureModal() {
+    if (modal) return modal;
+    var overlay = document.createElement('div');
+    overlay.style.cssText =
+      'position:fixed;inset:0;z-index:2147483647;background:rgba(0,0,0,0.55);' +
+      'display:none;align-items:center;justify-content:center;padding:1rem;';
+    var box = document.createElement('div');
+    box.style.cssText =
+      'background:#fff;border-radius:12px;width:min(960px,100%);height:min(90vh,720px);' +
+      'overflow:hidden;position:relative;display:flex;flex-direction:column;' +
+      'box-shadow:0 20px 60px rgba(0,0,0,0.4);';
+    var close = document.createElement('button');
+    close.type = 'button';
+    close.setAttribute('aria-label', 'Close');
+    close.innerHTML = '&times;';
+    close.style.cssText =
+      'position:absolute;top:0.5rem;right:0.5rem;width:34px;height:34px;border-radius:50%;' +
+      'border:0;background:rgba(0,0,0,0.06);color:#000;font-size:22px;line-height:1;' +
+      'cursor:pointer;z-index:2;';
+    close.addEventListener('click', closeModal);
+    var iframe = document.createElement('iframe');
+    iframe.style.cssText = 'flex:1;width:100%;border:0;display:block;';
+    iframe.setAttribute('allow', 'clipboard-write');
+    iframe.setAttribute('title', 'Booking');
+    box.appendChild(close);
+    box.appendChild(iframe);
+    overlay.appendChild(box);
+    overlay.addEventListener('click', function (ev) {
+      if (ev.target === overlay) closeModal();
+    });
+    document.addEventListener('keydown', function (ev) {
+      if (ev.key === 'Escape' && modal && modal.overlay.style.display !== 'none') closeModal();
+    });
+    document.body.appendChild(overlay);
+    modal = { overlay: overlay, iframe: iframe };
+    return modal;
+  }
+  function openModal(link, config) {
+    var m = ensureModal();
+    m.iframe.src = buildSrc(link, config);
+    m.overlay.style.display = 'flex';
+    document.body.style.overflow = 'hidden';
+  }
+  function closeModal() {
+    if (!modal) return;
+    modal.overlay.style.display = 'none';
+    modal.iframe.src = 'about:blank';
+    document.body.style.overflow = '';
+  }
+
+  function calendarIcon() {
+    var svg = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
+    svg.setAttribute('width', '18');
+    svg.setAttribute('height', '18');
+    svg.setAttribute('viewBox', '0 0 24 24');
+    svg.setAttribute('fill', 'none');
+    svg.setAttribute('stroke', 'currentColor');
+    svg.setAttribute('stroke-width', '2');
+    svg.setAttribute('stroke-linecap', 'round');
+    svg.setAttribute('stroke-linejoin', 'round');
+    svg.innerHTML =
+      '<rect x="3" y="4" width="18" height="18" rx="2" ry="2"></rect>' +
+      '<line x1="16" y1="2" x2="16" y2="6"></line>' +
+      '<line x1="8" y1="2" x2="8" y2="6"></line>' +
+      '<line x1="3" y1="10" x2="21" y2="10"></line>';
+    return svg;
+  }
+
+  var Calrs = {
+    inline: function (opts) {
+      opts = opts || {};
+      var target = typeof opts.selector === 'string' ? document.querySelector(opts.selector) : opts.selector;
+      if (!target) return;
+      var iframe = document.createElement('iframe');
+      iframe.src = buildSrc(opts.link, opts.config);
+      iframe.style.cssText = 'width:100%;min-height:560px;border:0;display:block;';
+      iframe.setAttribute('loading', 'lazy');
+      iframe.setAttribute('title', 'Booking');
+      target.innerHTML = '';
+      target.appendChild(iframe);
+      attachResize(iframe);
+    },
+
+    floatingButton: function (opts) {
+      opts = opts || {};
+      var btn = document.createElement('button');
+      btn.type = 'button';
+      var pos = opts.buttonPosition || 'bottom-right';
+      var bg = opts.buttonColor || '#2563eb';
+      var fg = opts.textColor || '#ffffff';
+      var positions = {
+        'bottom-right': 'bottom:1rem;right:1rem;',
+        'bottom-left':  'bottom:1rem;left:1rem;',
+        'top-right':    'top:1rem;right:1rem;',
+        'top-left':     'top:1rem;left:1rem;'
+      };
+      btn.style.cssText =
+        'position:fixed;' + (positions[pos] || positions['bottom-right']) +
+        'z-index:2147483646;padding:0.75rem 1.25rem;border:0;border-radius:999px;' +
+        'font-family:inherit;font-size:0.95rem;font-weight:600;cursor:pointer;' +
+        'display:inline-flex;align-items:center;gap:0.5rem;' +
+        'box-shadow:0 4px 14px rgba(0,0,0,0.15);' +
+        'background:' + bg + ';color:' + fg + ';';
+      if (opts.showIcon !== false) btn.appendChild(calendarIcon());
+      var label = document.createElement('span');
+      label.textContent = opts.buttonText || 'Book a meeting';
+      btn.appendChild(label);
+      btn.addEventListener('click', function () { openModal(opts.link, opts.config); });
+      document.body.appendChild(btn);
+    },
+
+    elementClick: function () {
+      document.addEventListener('click', function (ev) {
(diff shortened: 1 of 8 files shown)


src/web/mod.rs
```diff
@@ -877,16 +1001,26 @@ async fn csp_middleware(
     next: axum::middleware::Next,
 ) -> axum::response::Response {
     let booking_page = is_booking_form_path(request.uri().path());
+    let embed_mode = is_embed_request(request.uri().query());
```
olivierlambert: **Blocking**: the embed detection here is keyed on the query string alone, so it applies to every route. `GET /auth/login?embed=1` gets `frame-ancestors *`, and modern browsers prefer CSP `frame-ancestors` over the `X-Frame-Options: SAMEORIGIN` layer (CSP Level 2), so the login page becomes frameable by any site. Combined with the `SameSite=None` CSRF cookie (same query-only check in `csrf_cookie_middleware`), the framed login form is fully functional inside an attacker's iframe: login clickjacking / login CSRF.

Please gate both this swap and `cross_site_cookie` in `csrf_cookie_middleware` on something like `is_embeddable_path()` that matches only the public booking surfaces (`/u/...`, `/team/...`, `/g/...`, the legacy `/{slug}` pages and their `/book` children) and never `/dashboard` or `/auth`. A unit test asserting that `/auth/login?embed=1` keeps `frame-ancestors 'self'` would lock the fix in.

src/web/mod.rs
```diff
@@ -118,10 +118,22 @@ pub const CSRF_COOKIE_NAME: &str = "__Host-calrs_csrf";
...
+/// double-submit token-vs-cookie comparison still defends against CSRF — we
+/// never relied on SameSite for that. Browsers blocking all third-party
+/// cookies (Safari ITP, Chrome with 3P cookies disabled) still won't deliver
+/// it; that's an upstream-platform limitation outside our control.
 pub fn csrf_cookie_value(token: &str) -> String {
+    csrf_cookie_value_for(token, false)
+}
+
+pub fn csrf_cookie_value_for(token: &str, cross_site: bool) -> String {
+    let same_site = if cross_site { "None" } else { "Lax" };
```
olivierlambert, line 133: When `cross_site` is true, please also append `Partitioned` (CHIPS). Browsers that block unpartitioned third-party cookies (Safari 18.4+, Firefox 131+, Chrome profiles with third-party cookies disabled) will then still deliver the cookie inside the iframe, which is exactly the embed case. Without it, embedded bookings fail the CSRF check in those browsers. Worth extending `csrf_cookie_embed_uses_samesite_none` to assert the attribute too.

src/web/mod.rs
```diff
@@ -7042,6 +7180,77 @@ async fn delete_override(
...
+    let username: String = user.username.clone().unwrap_or_else(|| user.id.clone());
+    let cal_path = format!("/u/{}/{}", username, et_slug);
+
+    let base_url = std::env::var("CALRS_BASE_URL").unwrap_or_default();
+    let base_url = base_url.trim_end_matches('/').to_string();
+
+    // Pre-fill brand color from the org theme so the default snippet matches
+    // the rest of the site.
+    let accent: Option<String> =
+        sqlx::query_scalar("SELECT custom_accent FROM auth_config WHERE id = 'singleton'")
```
olivierlambert: This reads `custom_accent` unconditionally, but that column is only meaningful when `theme = 'custom'` (see `build_theme_css`). On a preset theme the prefill is stale or falls back to `#2563eb` and won't match the real org accent. Consider resolving the accent the same way `build_theme_css` does: preset accent when the theme is a preset, `custom_accent` otherwise.

src/web/mod.rs
```diff
@@ -7042,6 +7180,77 @@ async fn delete_override(
...
+    .await
+    .unwrap_or(None);
+
+    let (_et_id, et_slug, et_title, default_calendar_view, visibility) = match et {
+        Some(e) => e,
+        None => return Redirect::to("/dashboard/event-types").into_response(),
+    };
+
+    // Username for the cal link path: /u/{username}/{slug}.
+    let username: String = user.username.clone().unwrap_or_else(|| user.id.clone());
```
olivierlambert: `users.username` is nullable (migration 003) and all `/u/{username}` lookups are `WHERE username = ?`, so falling back to `user.id` generates a snippet whose link 404s with "User not found." Better to detect the missing username and render the page with a notice ("set a username in settings to enable embedding") instead of producing a dead snippet.

src/web/mod.rs
```diff
@@ -9827,9 +10057,12 @@ async fn show_slots_for_user(
     headers: HeaderMap,
     Path((username, slug)): Path<(String, String)>,
     Query(query): Query<SlotsQuery>,
-) -> impl IntoResponse {
+) -> Response {
+    let embed = query.embed_params();
     if username.contains('+') {
-        return show_dynamic_group_slots(&state, &headers, &username, &slug, &query).await;
+        return show_dynamic_group_slots(&state, &headers, &username, &slug, &query)
```
olivierlambert, line 10256: Nit: dynamic group pages (`alice+bob`) return here before the embed threading, so embedding those links shows full chrome and never autosizes. Fine to leave out of scope, but a short code comment saying so would make it read as intentional.

templates/base.html
```diff
@@ -427,6 +435,25 @@
...
+    body { padding: 1rem 0.75rem; background: transparent; display: block; min-height: 0; }
+    .container { max-width: none; }
+    .logo, .powered, .theme-pill, #theme-toggle { display: none !important; }
+    {% if embed_brand %}
+    /* Apply the brand accent in both light and dark mode. The `html.dark`
+       selector is needed to out-specify base.html's `html.dark` accent rule;
+       `:root` alone (lower specificity) would lose to it in dark mode. */
+    :root, html.dark {
+      --accent: {{ embed_brand }};
+      --accent-hover: {{ embed_brand }};
```
olivierlambert: Nit: only `--accent` and `--accent-hover` are overridden. `--accent-subtle`, `--accent-border` and `--accent-muted` (set by `/accent.css`) keep the org color, so a custom brand renders half-applied on slot pills and borders. Deriving the rgba variants from the brand hex here would complete the theming.

assets/embed.js
```diff
@@ -0,0 +1,184 @@
...
+ * | "dark"), and `brand` (hex color, with or without leading "#"). These end
+ * up as query parameters on the iframe src and feed into the embed-mode
+ * rendering on the calrs side.
+ */
+(function () {
+  if (window.Calrs) return;
+
+  function buildSrc(link, config) {
+    if (!link) return 'about:blank';
+    var url = link + (link.indexOf('?') >= 0 ? '&' : '?') + 'embed=1';
```
olivierlambert, line 24: Nit: if `link` contains a `#fragment`, the query gets appended after the fragment and is lost. Stripping the fragment first (or documenting that links must not carry one) avoids a confusing failure.

assets/embed.js
```diff
@@ -0,0 +1,184 @@
...
+        'background:' + bg + ';color:' + fg + ';';
+      if (opts.showIcon !== false) btn.appendChild(calendarIcon());
+      var label = document.createElement('span');
+      label.textContent = opts.buttonText || 'Book a meeting';
+      btn.appendChild(label);
+      btn.addEventListener('click', function () { openModal(opts.link, opts.config); });
+      document.body.appendChild(btn);
+    },
+
+    elementClick: function () {
```
olivierlambert, line 170: Nit: `elementClick()` is auto-invoked on load and also exposed on the public API, so a host page that calls it manually binds the document listener twice. A `bound` flag guard would make it idempotent.

olivierlambert (changes requested):
Thanks Arthur, this is a really solid contribution. The embed runtime is clean, the brand color validation and the postMessage origin + source checks are exactly right, and the resize feedback loop fix (with the comment explaining why body, not documentElement) is appreciated. CI is green and the new tests pin the right contracts.

One security issue blocks merge, plus a few smaller fixes (all inline). To settle the open design questions on our side:

- **Embedding policy**: we keep `frame-ancestors *` on booking pages, no origin allowlist. Matches Cal.com, and embedding is the point of the feature.
- **Third-party cookie blocking**: rather than just documenting the limitation, please add the `Partitioned` (CHIPS) attribute to the embed CSRF cookie. Details inline.
- **Team event types**: fine to keep this PR personal-only. We'll open a follow-up issue for a team embed generator, the underlying `/team/` pages already accept the embed params.

The blocking issue in one line: the `frame-ancestors` swap and the `SameSite=None` cookie are keyed on the query string alone, so `/auth/login?embed=1` becomes frameable and fully functional inside a third-party iframe. The fix is to gate both on the public booking paths. See the inline comment on `csp_middleware`.

src/web/mod.rs
arthur-perrot: Fixed - Added is_embeddable_path() — an allowlist of the public booking surfaces (/u/, /team/, /g/, plus legacy /{slug} and /{slug}/book guarded by a reserved-segment denylist). Both the frame-ancestors * swap and the SameSite=None CSRF cookie are now gated on is_embedded_booking_request(path, query). Verified at runtime: /auth/login?embed=1 and /dashboard?embed=1 keep frame-ancestors 'self' and a Lax cookie; only booking surfaces get *. Locked in with is_embeddable_path_allows_booking_surfaces_only and embedded_booking_request_requires_both_flag_and_path.

src/web/mod.rs
arthur-perrot: csrf_cookie_value_for now appends ; Partitioned when cross_site is true. Extended csrf_cookie_embed_uses_samesite_none to assert it, and added csrf_cookie_non_embed_is_samesite_lax_unpartitioned to assert the non-embed cookie stays Lax and unpartitioned.

src/web/mod.rs
arthur-perrot: Accent prefill now resolves with the same precedence as build_theme_css: the preset's accent for a preset theme (new preset_accent() helper), custom_accent only when theme = 'custom', default blue otherwise.

src/web/mod.rs
arthur-perrot: embed_page detects a missing username and renders a "set a username in settings" notice instead of a dead snippet (has_username flag; the snippet/preview block is skipped).

src/web/mod.rs
arthur-perrot: documented - Added a comment marking the early return as intentionally out of scope (the generator only emits single-user/team links).

templates/base.html
arthur-perrot: Fixed - brand_css() now derives --accent-subtle/--accent-border/--accent-muted from the brand hex via a new hex_to_rgb() helper, so slot pills and borders pick up the brand too. Covered by hex_to_rgb_parses_3_and_6_digit and brand_css_derives_rgba_variants.

assets/embed.js
arthur-perrot: fixed - buildSrc() strips the fragment before appending ?embed=1&….

assets/embed.js
arthur-perrot: fixed - Added an elementClickBound guard so manual + auto invocation is idempotent.

arthur-perrot:
Corrections fixed
#126 Team embed generator created in commit 1fd713c

olivierlambert (commented):
Re-reviewed everything, great turnaround. All eight points are properly addressed:

- Verified the `is_embeddable_path()` denylist against every top-level route in both routers (`auth`, `dashboard`, `booking`, `t`, all asset routes): complete. The token URLs (`/booking/cancel/{token}` etc.) are also excluded by the segment-shape check. The two regression tests pin exactly the right contract.
- Confirmed at runtime on the branch: `/auth/login?embed=1` and `/dashboard?embed=1` keep `frame-ancestors 'self'` + `SameSite=Lax`, while `/u/{user}/{slug}?embed=1` and legacy `/{slug}?embed=1` get `frame-ancestors *` + `SameSite=None; Partitioned`.
- `hex_to_rgb` / `brand_css` derivation checked, including the injection-attempt test. Output stays HTML-escape-safe.
- Happy to take the team embed generator in this PR rather than as a follow-up; the access gating (team admins + global admins) matches the permission model from #55. I'll close #126 when this merges.

One last small thing before I approve, on the new team handler (inline comment): the private-team case generates a snippet that can't work without the team invite token.

src/web/mod.rs
```diff
@@ -7042,6 +7277,167 @@ async fn delete_override(
...
+) -> Response {
+    let user = &auth_user.user;
+    let is_admin = user.role == "admin";
+
+    if !is_admin && !is_team_admin(&state.pool, &user.id, &team_id).await {
+        return Html("You are not a team admin of this team.".to_string()).into_response();
+    }
+
+    let et: Option<(String, String, String, String, String)> = sqlx::query_as(
+        "SELECT et.slug, et.title, et.default_calendar_view, et.visibility, t.slug \
```
olivierlambert: This doesn't fetch `t.visibility`, so for a **private team** the generated snippet points at `/team/{slug}/{event}` without the required `?invite={invite_token}`, and embedded visitors get "Event type not found." The warning banner in `embed.html` only keys off the event type's visibility, not the team's.

Could you add `t.visibility` to this select and extend the warning, e.g. when the team is private, show the same kind of heads-up as for private/internal event types ("this team is private; embedded visitors can't book without an invite link")? That's the last item, I'll approve once it's in.

arthur-perrot:
### new behavior

- When the **team is private** and the **event is public**, only the new **"This team is private"** banner is displayed.
- When the **team is switched back to public**, no banner will be displayed.
- When the **event type is set to internal or private** while the **team remains public**, only the **"This event type is internal/private"** banner is displayed.
- When **both conditions are true** (**private team** + **private/internal event**), both banners are displayed, stacked one above the other.

src/web/mod.rs
arthur-perrot: normally done 🫡

olivierlambert (approved):
Verified the private-team banner: `t.visibility` threaded through `group_embed_page`, independent banner with an `is defined` guard so the personal embed page is untouched, and both warnings can stack when team and event type are each restricted — which is the correct behavior.

Full-PR recap of what was verified across the rounds: path-gated framing relaxation (runtime-checked: `/auth/login?embed=1` and `/dashboard?embed=1` stay locked down, only booking surfaces get `frame-ancestors *` + `SameSite=None; Partitioned`), RESERVED denylist still complete against every top-level route on current main, brand theming derives all five accent vars, username notice instead of dead snippets, and the embed.js fixes. 737 tests pass locally and CI ran green on the merged state.

Thanks for the quality work across this whole PR, including the team generator. Merging.

PR #127, opened 2026-06-05
fix(ews): timed UTC events, all-day date offset, and recurring series

arthur-perrot:
## Summary

Four correctness fixes in the EWS (Microsoft Exchange) sync path, all surfaced on an Exchange 2019 mailbox in `Europe/Paris`. After this PR, timed events block slots, all-day events land on the right local date, and recurring series sync as one row per occurrence.

### Timed UTC events silently dropped from busy-time calculation

EWS emits UTC timestamps as `YYYYMMDDTHHMMSSZ` (e.g. `20260606T081000Z`). The shared `parse_datetime` helper only accepted `%Y%m%dT%H%M%S` — chrono rejected the trailing `Z` with `ParseError(TooLong)`, so every EWS timed event was silently filtered out by `filter_map` before reaching `has_conflict`. Result: slots that should have been busy were displayed as available.

**Fix:** strip a trailing `Z` at the top of `parse_datetime` in all three locations (`src/web/mod.rs`, `src/commands/booking.rs`, `src/commands/event_type.rs`). Downstream `convert_event_to_tz` already reads the event's `timezone = "UTC"` column and applies the correct UTC→host conversion — DST-safe in both summer (CEST) and winter (CET), verified by paired regression tests.

### All-day events shifted to the previous local day

EWS reports all-day boundaries in UTC. A Paris all-day event for `13/06` arrives as `Start = 2026-06-12T22:00:00Z`. `format_dt` in `src/ews/ical.rs` did `value.chars().take(10)` and stored `20260612` — shifting the event one local day back.

**Fix:** drop the `VALUE=DATE` branch entirely; emit all-day items as timed UTC values like every other event. Downstream parse + `convert_event_to_tz` then resolves the correct local date for any host TZ.

### Recurring series completely missing

`CalendarView` expands recurring events into individual occurrences server-side, all sharing the master's UID. `synth_or_fetch_mime` routed any item with `<Recurrence>` to a follow-up `GetItem` MIME fetch — but on expanded virtual occurrence IDs Exchange frequently returns the metadata block without `MimeContent`, and `parse_get_item_response` silently skipped those, losing whole series.

**Fix:** drop the MIME round-trip. `CalendarView` already gives us every occurrence's start/end. Emit a `RECURRENCE-ID = occurrence start` so siblings sharing the master UID stay addressable in the `(calendar_id, uid, recurrence_id)` upsert key instead of collapsing into one row.

### Occurrence detection used the wrong XML signal

`has_recurrence` was driven by `block.contains("Recurrence>")`. Exchange 2019 does **not** include the master's `<t:Recurrence>` pattern on expanded occurrences returned by `CalendarView` — the heuristic always returned `false`, so even after the fix above no `RECURRENCE-ID` was emitted and every occurrence still collapsed.

**Fix:** request `calendar:CalendarItemType` in the `FindItem` payload and use it as the canonical signal — `Occurrence` / `Exception` flag the item. A raw `RecurringMaster` (only reachable via the unused non-windowed `fetch_events`) is intentionally NOT flagged — masters are a single logical item, and emitting `RECURRENCE-ID` without `RRULE` would be meaningless.

## Behaviour changes worth noting

- **CalDAV with compact UTC events**: rare, but if any of your CalDAV sources emit `YYYYMMDDTHHMMSSZ` (e.g. some Exchange-bridged stacks), those events were silently NOT blocking before and will now correctly block. End-users may perceive this as "new" conflicts that always should have been there.
- **EWS all-day storage format**: now 16-char UTC (`20260612T220000Z`) instead of 8-char date (`20260612`). The orphan sweep cleans up old single-row remnants on first sync — no manual data migration needed.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test --bin calrs` — 737/737 pass, 11 new regression tests
- [x] Live-tested against an Exchange 2019 mailbox: timed events block slots; all-day events show on the correct day; recurring series appears per-occurrence in the troubleshoot view

## Files changed

| File | Change |
| --- | --- |
| `src/web/mod.rs`, `src/commands/booking.rs`, `src/commands/event_type.rs` | `parse_datetime` strips trailing `Z` |
| `src/ews/ical.rs` | `format_dt` drops `VALUE=DATE`; `synth_vcalendar` emits `RECURRENCE-ID` for occurrences |
| `src/ews/mod.rs` | `synth_or_fetch_mime` drops the MIME round-trip |
| `src/ews/operations.rs` | Adds `calendar:CalendarItemType` to the `FindItem` request |
| `src/ews/parse.rs` | `has_recurrence` is now driven by `CalendarItemType` |


src/web/mod.rs
```diff
@@ -10583,6 +10583,10 @@ async fn handle_booking_for_user(
 // --- Slot computation (shared with CLI) ---
 
 fn parse_datetime(s: &str) -> Option<NaiveDateTime> {
+    // EWS UTC timestamps reach us as `YYYYMMDDTHHMMSSZ`; strip the trailing
+    // `Z` so the naive value parses and timezone-conversion (driven by the
+    // event's `timezone` column = "UTC") can do the proper UTC→host offset.
+    let s = s.strip_suffix('Z').unwrap_or(s);
```
olivierlambert: **Requested change:** this PR patches the same 4 lines into three separate copies of `parse_datetime` (here, `commands/booking.rs`, `commands/event_type.rs`), and the codebase already has a fourth copy that handled `Z` correctly all along: `rrule::parse_ical_datetime` (public, identical format list, strips the suffix). That drift between copies is the root cause of this very bug: the rrule path kept up, the busy paths didn't.

Please consolidate: either have the three `parse_datetime` bodies delegate to `crate::rrule::parse_ical_datetime`, or move that helper to `utils.rs` and call it from all four sites. The next format quirk Exchange invents should only need fixing once. Happy to approve as soon as that's in; everything else in the PR checks out.

src/ews/mod.rs
```diff
@@ -228,52 +228,28 @@ impl CalendarProvider for EwsProvider {
...
+/// Build a `RawEvent` for each item via [`ical::synth_vcalendar`].
+///
+/// We don't follow up with a MIME `GetItem` for recurring items: `CalendarView`
+/// already expanded every occurrence in the requested window, and for those
+/// virtual occurrence IDs Exchange frequently returns the metadata block
+/// without `MimeContent` — which the parser then silently drops, losing the
+/// entire series. Synthesising directly from the occurrence's own Start/End
+/// keeps every one, and the `RECURRENCE-ID` emitted by `synth_vcalendar`
+/// makes them addressable under their shared master UID.
 async fn synth_or_fetch_mime(
```
olivierlambert: Nit: with the MIME branch gone this name is a misnomer; it never fetches MIME, ignores `_provider`, and no longer needs `async`. `fn synth_raw_events(items: Vec<parse::EwsCalendarItem>) -> Vec<RawEvent>` would say what it does.

src/ews/ical.rs
```diff
@@ -62,6 +62,15 @@ pub fn synth_vcalendar(item: &EwsCalendarItem) -> Option<String> {
...
     buf.push_str(&format!("DTSTART{dtstart}\r\n"));
     buf.push_str(&format!("DTEND{dtend}\r\n"));
+    if item.has_recurrence {
+        // CalendarView returns occurrences of a recurring series as separate
+        // items sharing the master's UID. Without RECURRENCE-ID they would
+        // collide on the `(calendar_id, uid, recurrence_id)` upsert key and
+        // collapse into one row, dropping every occurrence but the last.
+        // Using the occurrence's start as RECURRENCE-ID keeps each one
+        // addressable.
+        buf.push_str(&format!("RECURRENCE-ID{dtstart}\r\n"));
```
olivierlambert, line 79: Nit, no change needed: canonically RECURRENCE-ID is the occurrence's *original* start, and using the new start means a rescheduled exception moved exactly onto a sibling occurrence's start collides on the upsert key (one row wins). For busy-time purposes the impact is nil (same start, both block) and getting the original start out of EWS costs another property fetch, so the trade-off is right. A one-line comment acknowledging the edge would help the next reader.

olivierlambert (changes requested):
Another strong PR, thanks. I verified the load-bearing claims rather than taking them on faith, and they all hold up:

- `upsert_provider_events` stores `extract_vevent_tzid()`, which returns `UTC` for trailing-`Z` values, so the stripped naive value really does get the UTC→host conversion (for both EWS and CalDAV).
- The upsert key `(calendar_id, uid, COALESCE(recurrence_id, ''))` picks up the new `RECURRENCE-ID`, and `remove_orphaned_ews_events` matches on `(uid, recurrence_id)`, so the pre-fix collapsed series row is swept on first sync as described. The Z-suffixed `start_at` still sorts correctly against the `YYYYMMDD` window prefix.
- Production EWS sync only ever calls `fetch_events_since` → `CalendarView`, so `RecurringMaster` is indeed unreachable and dropping the MIME round-trip can't lose RRULE fidelity on a live path.
- The DST round-trip pair and the all-day reasoning check out; the old `VALUE=DATE` path stored a floating date on the UTC-shifted wrong day, the new form converts to exactly local midnight-to-midnight. Strictly better.
- The CalDAV behavior change (Z-suffixed events now block) is the correct outcome; those conflicts were always real.

737/737 pass locally. One substantive change requested before I approve (inline), plus two optional nits.

src/ews/ical.rs
arthur-perrot: Added a comment in src/ews/ical.rs

src/ews/mod.rs
arthur-perrot: Renamed to synth_raw_events, dropped the _provider argument and the async qualifier; updated both call sites in EwsProvider::fetch_events and fetch_events_since

src/web/mod.rs
arthur-perrot: done, moved to utils and updated all calls

src/lib.rs
```diff
@@ -0,0 +1,10 @@
+// Temporary library shim so examples (e.g. `examples/ews_smoke.rs`) can
+// exercise internal modules. Not exported in published crates.
+#![allow(dead_code)]
```
olivierlambert: **Requested:** please drop `src/lib.rs` and move the smoke test into `src/ews/` as an `#[ignore]`d test instead, e.g.:

    #[tokio::test]
    #[ignore = "needs a real Exchange server; set EWS_URL/EWS_USER/EWS_PASS"]
    async fn ews_smoke() { ... }

Run with `cargo test -- --ignored ews_smoke` and the same env vars. Same tool, zero structural change: no dual compilation, no duplicated test runs, no blanket `dead_code` allow, and it lives next to the code it exercises. The example's nice step-by-step output can stay as `println!`s — `cargo test -- --ignored --nocapture` shows them.

With that, this PR is ready; everything else is verified and approved in spirit.

olivierlambert (commented):
Verified the fix round:

- Consolidation is exactly right: one `parse_ical_datetime` in `utils.rs`, with `web/mod.rs`, `commands/booking.rs`, `commands/event_type.rs` and `rrule.rs` all delegating. The four drifting copies are gone.
- `synth_raw_events` rename + signature cleanup, and the RECURRENCE-ID caveat comment, both good.

One remaining ask before approval, on the unrequested `src/lib.rs` + `examples/ews_smoke.rs` addition (inline comment). The smoke test itself is genuinely useful — Exchange can't be unit-tested and a repeatable end-to-end probe has real value — but the library shim it rides on has costs I measured rather than guessed:

- `main.rs` still declares `mod caldav/ews/providers/utils`, so those four modules now compile twice (lib target + bin target), and all 110 of their unit tests run twice (`synth_all_day*` runs 2+2, `parse_ical*` runs 7+7 across the two targets). That slows every build, every CI run, and the pre-commit hook.
- The blanket `#![allow(dead_code)]` conflicts with the project convention that dead_code warnings stay visible (see CLAUDE.md, Known compiler warnings).
- `calrs::ews::X` and the bin's `crate::ews::X` are now distinct types with identical names, which is a foot-gun for future contributors.

arthur-perrot:
All good !

```
cargo test --bin calrs ews::
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.31s
     Running unittests src/main.rs (target/debug/deps/calrs-004f86eee79cb3ee)

running 34 tests
test ews::autodiscover::tests::no_match_returns_none ... ok
test ews::autodiscover::tests::conventional_url_format ... ok
test ews::autodiscover::tests::parse_modern_autodiscover_response ... ok
test ews::ical::tests::synth_emits_recurrence_id_for_recurring_occurrence ... ok
test ews::ical::tests::synth_basic_event ... ok
test ews::ical::tests::synth_all_day_utc_keeps_z_suffix_for_local_conversion ... ok
test ews::ical::tests::synth_all_day_naive_passes_through_floating ... ok
test ews::ical::tests::synth_marks_free_as_transparent ... ok
test ews::ical::tests::synth_omits_recurrence_id_for_non_recurring_event ... ok
test ews::autodiscover::tests::fallback_to_asurl ... ok
test ews::ical::tests::synth_marks_cancelled_status ... ok
test ews::ical::tests::synth_uses_item_id_when_uid_missing ... ok
test ews::parse::tests::extract_vcalendar_from_mime ... ok
test ews::parse::tests::parses_create_item_response_returns_id ... ok
test ews::parse::tests::count_response_messages_works ... ok
test ews::parse::tests::parses_find_folder_response ... ok
test ews::parse::tests::occurrence_calendar_item_type_marks_has_recurrence ... ok
test ews::parse::tests::single_calendar_item_type_does_not_mark_has_recurrence ... ok
test ews::parse::tests::parses_find_item_calendar_view ... ok
test ews::soap::tests::attr_extracts_value ... ok
test ews::parse::tests::parses_sync_delta_with_create_update_delete ... ok
test ews::soap::tests::collect_blocks_returns_each_record ... ok
test ews::soap::tests::escape_special_chars ... ok
test ews::soap::tests::extract_returns_none_for_self_closing ... ok
test ews::soap::tests::extract_simple_tag_content ... ok
test ews::soap::tests::extract_with_attributes ... ok
test ews::soap::tests::fault_detection ... ok
test ews::soap::tests::fault_from_response_code ... ok
test ews::soap::tests::fault_no_error_returns_none ... ok
test ews::soap::tests::unescape_roundtrip ... ok
test ews::tests::ews_provider_trims_trailing_slash ... ok
test ews::tests::ews_smoke ... ignored, needs a real Exchange server; set EWS_URL/EWS_USER/EWS_PASS
test ews::tests::upper_bound_extends_two_years ... ok
test ews::tests::upper_bound_falls_back_for_garbage_input ... ok

test result: ok. 33 passed; 0 failed; 1 ignored; 0 measured; 693 filtered out; finished in 0.01s
```

olivierlambert (approved):
Verified the final round: `lib.rs` and `examples/` removed, smoke test converted to an `#[ignore]`d test in `src/ews/mod.rs` (nice touch using `try_init()` so it can't panic on double subscriber init). Back to a single test target — 726 pass + 1 ignored locally, no more duplicated module tests, and the dead_code convention is intact.

With the earlier rounds (Z-suffix parse consolidated into one `utils::parse_ical_datetime`, `synth_raw_events` cleanup, RECURRENCE-ID caveat documented), this is ready. Approved.

One micro-nit I'll fix on main after merge, not worth another round: the doc comment's run command says `cargo test -p calrs --lib ews::tests::ews_smoke` — with the lib target gone, `--lib` now errors; it should be plain `cargo test ews_smoke -- --ignored --nocapture`.

Issue #45, opened 2026-04-23
feat: auto-generate video call links per booking (Zoom, Meet, Jitsi, etc.)


gounthar:
I'm passing this along from a video review of cal.rs by Nicolas Vivant:
- Video: https://lnkd.in/emEJYg4D
- LinkedIn post: https://www.linkedin.com/posts/vivant_logicielslibres-opensource-freesofwtare-share-7452697384326791168-s88N

I'm not a power user of scheduling tools, so take my framing with some salt. But I didn't want this observation to get lost.

Around 10:33 in the video, Nicolas points out that when setting up an event type, the video call field only accepts a static URL you paste in manually. He thinks auto-generation will probably arrive at some point, and he acknowledges it's a known gap vs. Calcom or Calendly, which generate a fresh unique link per confirmed booking (Zoom, Google Meet, Jitsi, etc.).

The practical issue: with a static URL, the same room link gets reused for every booking. That may be fine for some workflows. For others (coaching sessions, anything where you want isolation between bookings) it's a problem.

If this is already on the roadmap, sorry for the noise. Close it and I'll get out of the way.

olivierlambert:
Hi @gounthar, this is not noise at all. You are right that it is a real gap, and you are also right that it is going to arrive at some point. Thanks for relaying it.

The reason I have not done it yet is that "auto-generate video link per booking" is not one feature, it is a family with very different costs depending on the provider:

- **Jitsi (public meet.jit.si)**: no API, no OAuth, just generate a unique room name per booking. Roughly half a day of work.
- **Jitsi self-hosted with JWT**: sign a JWT per booking with a configured shared secret. Roughly a day.
- **Generic webhook**: configure a URL that calrs POSTs to on confirm, expect a meeting URL back. Covers any custom or self-hosted provider. About a day.
- **Google Meet**: requires OAuth against Google Calendar API, encrypted refresh-token storage, calendar event creation with `conferenceData`. Several days, plus per-user setup.
- **Zoom**: register a Zoom Marketplace app, OAuth, Create Meeting API, encrypted token storage. Several days.
- **Microsoft Teams**: Microsoft Graph API, similar OAuth flow to Google. Several days.

Phased plan:

1. Ship Jitsi unique-room-per-booking first. Quick win, no external dependencies, demonstrates the pattern end to end (DB, email, ICS).
2. Add the generic webhook option, which lets self-hosters bring their own provider without me writing per-vendor integrations.
[...]

PR #128, opened 2026-06-05
feat(meeting): auto-generate video meeting links per booking

arthur-perrot:
WIP from issue #45.

Auto-generated video meeting links per booking, shipped as phases 1 + 2 of the plan you laid out in [this comment](https://github.com/olivierlambert/calrs/issues/45#issuecomment-4322352688):

- ✅ **Phase 1 — Jitsi unique-room-per-booking.** Local URL builder, no external dependency.
- ✅ **Phase 2 — Generic webhook.** Lets self-hosters bring their own provider (Jitsi+JWT, Whereby, a custom Meet, anything that can return a URL).
- ⏭️ Phase 3 (Zoom / Google Meet / Teams OAuth) deliberately not included — multi-day per provider, out of scope here. Can land in follow-up PRs.

### Jitsi (auto-generated room)

A new `Jitsi (auto-generated room)` option appears in the event-type location dropdown. When picked, every confirmed booking gets a fresh URL of the form `<base_url>/<expanded-pattern>`.

Pattern tokens (WordPress-style, combinable in any order, unknown tokens are kept verbatim so typos are loud):

| Token | Expands to |
|---|---|
| `{username}` | host's username |
| `{event}` | event type slug |
| `{date}` | `YYYYMMDD` of the meeting start |
| `{random}` | 8 cryptographic random alphanumerics |

Default pattern is `{event}-{random}` (mirrors what cal.com does). The pattern is configurable org-wide in the admin panel, and can be overridden per event type.

### Generic webhook (bring-your-own provider)

A new `Webhook (custom provider)` option. On booking confirmation calrs POSTs:

```json
{
  "booking_uid": "…@calrs",
  "event_slug": "intro",
  "host_username": "alice",
  "guest_name": "Bob",
  "guest_email": "bob@example.com",
  "start_at": "2026-06-06T14:00:00",
  "end_at":   "2026-06-06T14:30:00"
}
```

Receiver returns `{"url": "https://…"}`. Optional HMAC-SHA256 signing — when enabled, calrs sends `X-Calrs-Signature: sha256=<hex>` over the raw body. The shared secret is encrypted at rest with the same AES-256-GCM machinery as the OIDC client secret.

### Admin panel — `/dashboard/admin`

Two new sections following the existing OIDC/Captcha pattern (CSRF, keep-current-secret on empty submit, cached config in `AppState` so guest pages don't hit `auth_config` on every render):

- **Jitsi**: base URL, room pattern, optional display name (e.g. `Meet DYB`)
- **Meeting webhook**: URL, auth mode (`none` / `hmac`), shared secret, optional display name

The display name is what guests see on the slot picker and booking form instead of the generic "Video call" badge. Empty falls back to the generic label so existing deployments are unaffected.

### Storage (migration 056)

```
auth_config:
  + jitsi_base_url, jitsi_pattern, jitsi_display_name
  + meeting_webhook_url, meeting_webhook_auth_mode,
    meeting_webhook_secret (encrypted), meeting_webhook_display_name

event_types:
  + meeting_pattern_override   -- optional per-event-type override of jitsi_pattern

bookings:
  + meeting_url                -- the actual generated URL for this booking
```

`bookings.meeting_url` is the source of truth once the booking is confirmed — emails, ICS attachment, and CalDAV write-back all read from it so they can never diverge (which would otherwise happen with `{random}`).

### Generation lifecycle

- Auto-confirm booking → URL generated and persisted immediately, then used in the confirmation email, ICS, and CalDAV push.
- Pending booking (requires confirmation) → generation deferred to `/booking/approve/{token}`. A host who **declines** never triggers a configured webhook.
- Wired into all four booking insertion paths (personal, team round-robin, dynamic group, legacy single-tenant) + the approve-by-token path via a single `resolve_booking_location` helper.

## Test plan

- [x] `cargo build`
- [x] `cargo clippy -- -D warnings`
- [x] `cargo fmt --check`
- [x] `cargo test` — 837 tests passing, including new unit tests for pattern expansion, RFC 4231 HMAC-SHA256 test vector, URL builder (override + default fallback), webhook auth-mode round-trip, and date extraction.
- [x] **End-to-end with curl** on an ephemeral data dir:
  - Registered first user (auto-admin), saved Jitsi config with `display_name=Meet DYB` and `pattern={event}-{date}-{random}`.
  - Created an event type with `location_type=jitsi_auto`.
  - Slot page renders the **Meet DYB** badge (not "Video call").
  - Booked two slots back-to-back → two distinct URLs (`…intro-20260606-ib35cpcq` / `…intro-20260606-seaptgry`) — the per-booking isolation use case from the original issue.
  - Configured webhook with HMAC + secret, pointed at a Python stub on `:13599`. Booked → calrs POSTed the JSON payload with `X-Calrs-Signature`. Verified the HMAC manually with `openssl dgst -sha256 -hmac` — exact match. Stub returned a custom URL, which was persisted to `bookings.meeting_url` and rendered on the confirmation page as a clickable link.
- [x] UI/UX test to link our self-hosted jitsi, create events, received good link inside calendar

image

## Out of scope

- Phase 3 OAuth providers (Zoom, Google Meet, Microsoft Teams). The webhook path already lets self-hosters bridge these via a small adapter.
- i18n: new admin labels follow the existing untranslated pattern of the OIDC/Captcha sections. The reused `confirmed-detail-location` Fluent key already covers the post-booking confirmation page for the new providers. Can be corrected if functionality is valid


src/web/meeting.rs
```diff
@@ -0,0 +1,618 @@
...
+                guest_email,
+                start_at: &start_at,
+                end_at: &end_at,
+            };
+            call_webhook(webhook_cfg, &payload).await.ok()
+        }
+        _ => None,
+    }?;
+
+    let _ = sqlx::query("UPDATE bookings SET meeting_url = ? WHERE id = ?")
```
olivierlambert, line 459: **Requested:** make this function idempotent — read `meeting_url` first and return it if already set, only generating when NULL/empty. Two reasons: (a) it makes the column the actual source of truth the design describes; (b) if two confirm paths ever race (dashboard approve + email token), the `{random}` room can't rotate between the email and the ICS.

This is also the natural place to fix the three missing consumers, which currently read `et.location_value` (empty for auto providers):
- `run_reminder_loop` — reminder emails for auto-provider bookings carry no meeting link today
- `confirm_booking` (dashboard approve) — never generates a URL at all; only `approve_booking_by_token` was wired
- `guest_reschedule_booking` — emails + CalDAV re-push lose the link after a reschedule

`COALESCE(NULLIF(b.meeting_url,''), et.location_value)` in those queries plus wiring `confirm_booking` through `resolve_booking_location` covers all three. A regression test asserting the reminder query picks up `meeting_url` would lock it in.

src/web/meeting.rs
```diff
@@ -0,0 +1,618 @@
...
+
+    let parsed: WebhookResponse = match resp.json().await {
+        Ok(p) => p,
+        Err(e) => {
+            tracing::warn!(error = %e, "meeting webhook response parse failed");
+            return Err(());
+        }
+    };
+
+    let url = parsed.url.trim().to_string();
```
olivierlambert, line 334: **Requested:** validate the scheme here — reject anything that doesn't start with `http://` or `https://`. The returned value is rendered as `<a href="...">` on the confirmation page and embedded in emails/ICS, so a compromised webhook receiver returning `javascript:...` would otherwise get a click-to-XSS on the guest. One line of defense-in-depth.

src/web/meeting.rs
```diff
@@ -0,0 +1,618 @@
...
+        LOCATION_TYPE_WEBHOOK => cfg.webhook.as_ref().and_then(|w| w.display_name.clone()),
+        _ => None,
+    }
+}
+
+/// Expand `{username}`, `{event}`, `{date}`, `{random}` in `pattern`.
+///
+/// Unknown placeholders are kept verbatim (e.g. `{foo}` stays `{foo}`) so a
+/// typo in the admin panel is loud rather than silently swallowed.
+pub fn expand_pattern(pattern: &str, tokens: &PatternTokens<'_>) -> String {
```
olivierlambert, line 208: Sanitizing the token *values* but not the pattern's literal characters leaves a gap: `meeting_pattern_override` is settable by any user on their own event types, so spaces, `?`, `#` or unicode in the pattern flow straight into the URL and can break the link in ICS/email contexts. Simpler and safer to run the final expanded room string through `sanitize_for_url`-style filtering (keeping `{`/`}` handling as-is), rather than each token individually.

templates/admin.html
```diff
@@ -372,6 +372,75 @@ <h2>Captcha</h2>
...
+      <input type="url" name="jitsi_base_url" value="{{ jitsi_base_url }}" placeholder="https://meet.example.com">
+    </div>
+    <div class="form-group">
+      <label>Display name <span class="hint">(optional)</span></label>
+      <input type="text" name="jitsi_display_name" value="{{ jitsi_display_name }}" placeholder="e.g. Meet DYB">
+      <span style="font-size: 0.8rem; color: var(--text-muted);">Shown to guests on the slot picker and booking form. Defaults to "Video call" if empty.</span>
+    </div>
+    <div class="form-group">
+      <label>Room name pattern</label>
+      <input type="text" name="jitsi_pattern" value="{{ jitsi_pattern }}" placeholder="{{ jitsi_default_pattern }}">
```
olivierlambert: Please add a warning hint (here and on the per-event-type override field) when the pattern lacks `{random}`: without it, rooms are predictable and two bookings of the same event type on the same day share a room, so one guest can sit in another's meeting. Stable rooms stay allowed — admins may want a personal-room-per-host setup — but the trade-off should be visible at configuration time, not discovered by a guest.

olivierlambert (changes requested):
Thanks, this is a well-built implementation of the issue #45 plan and the module itself is in great shape: the AppState caching follows the captcha/OIDC pattern, the secret handling (AES-256-GCM at rest, keep-on-empty submit) matches the existing machinery, migration 056 is registered in db.rs with the count tests updated, the RFC 4231 vector test is appreciated, and the lifecycle intent (defer generation for pending, never fire the webhook on decline) is exactly right. 741 tests pass locally.

One design-level problem blocks merge, though: **`bookings.meeting_url` is write-only.** The PR description calls it "the source of truth — emails, ICS attachment, and CalDAV write-back all read from it", but nothing in the codebase reads the column; the initial confirmation flow only works because the *return value* of `generate_and_persist` is threaded into `BookingDetails`. Three consumers still read `et.location_value`, which is empty for auto providers:

1. **Reminder emails** — `run_reminder_loop`'s query selects `et.location_value`, so reminders for Jitsi/webhook bookings go out with **no meeting link**. That's the email guests open right before the meeting.
2. **Dashboard approve** — only the email-token path (`approve_booking_by_token`) was wired to `resolve_booking_location`. Approving the same pending booking from the dashboard (`confirm_booking`) never generates a URL at all: confirmation email, ICS and CalDAV push go out link-less and nothing retries.
3. **Guest reschedule** — `guest_reschedule_booking` builds its emails and the CalDAV re-push from `et.location_value`, so the link is dropped after a reschedule even though `meeting_url` still sits in the DB.

The fix is mechanical: `COALESCE(NULLIF(b.meeting_url,''), et.location_value)` in those three queries, wire `confirm_booking` the same way as the token path, and make `generate_and_persist` idempotent (details inline).

Two smaller requested changes (inline): scheme-validate the webhook response URL, and sanitize the expanded room rather than only the token values. Plus one UI ask on `{random}`.

Decisions from our side so you don't have to guess:
- SSRF posture is fine as-is: the webhook URL is admin-only and localhost adapters are the primary use case, so the CalDAV private-host guard intentionally doesn't apply. A sentence in the docs would be welcome.
- Patterns without `{random}` stay allowed (stable rooms are a legitimate choice) but should warn in the UI.

arthur-perrot:
All review comments addressed. Summary of changes:

`src/web/meeting.rs
`

- generate_and_persist is now idempotent: reads bookings.meeting_url first and short-circuits if set, so two confirm paths racing can't rotate the {random} room between the email body and the ICS attachment.
- call_webhook validates the response URL's scheme — only http(s) accepted, defending against javascript: click-to-XSS via a compromised receiver.
- expand_pattern now runs the fully expanded room through a final sanitizer (sanitize_room, which preserves {} so unknown placeholders survive verbatim), instead of sanitizing individual token values. Closes the gap where pattern literals (?, #, spaces, unicode) leaked into URLs.
- Added module-level SSRF posture note.
- Three new tests: expand_pattern_sanitizes_pattern_literal, reminder_query_prefers_meeting_url_over_location_value, reminder_query_falls_back_to_location_value, generate_and_persist_is_idempotent. Renamed the sanitization test to match the new function. Total: 748 tests passing.

`src/web/mod.rs
`

- run_reminder_loop query: et.location_value → COALESCE(NULLIF(b.meeting_url, ''), et.location_value) so reminders carry the auto-provider link.
- confirm_booking (dashboard approve): same COALESCE in the query and wired through resolve_booking_location so the email/ICS/CalDAV push carry a real link for auto providers.
- guest_reschedule_booking: same COALESCE so the link survives a reschedule.

`templates/admin.html
`

- Live warning under the Jitsi pattern field when {random} is absent (checks the effective pattern, default-aware).

`templates/event_type_form.html`

- Same warning on the per-event-type meeting_pattern_override field, only shown when the user typed an override that lacks {random} (empty falls back to the org default, already warned on the admin page).


`cargo test `(745 + 3 new = 748 passing), `cargo clippy -- -D warnings` clean, `cargo fmt --check` clean, manual end-to-end booking confirmed `bookings.meeting_url` is populated and the reminder COALESCE resolves to it.

olivierlambert (approved):
Verified the whole fix round on the branch:

- `generate_and_persist` idempotency: reads `meeting_url` first, short-circuits when set, covered by a real in-memory-DB test. The column is now genuinely the source of truth.
- All three missing consumers fixed: reminder loop, dashboard approve (`confirm_booking` now mirrors the token path through `resolve_booking_location`), and guest reschedule, each via `COALESCE(NULLIF(b.meeting_url,''), et.location_value)`.
- Scheme validation on the webhook response with the right reasoning in the comment.
- `sanitize_room` over the fully expanded pattern, preserving `{}` so typos stay loud; the traversal test survived the rename.
- `{random}` warnings on both the admin pattern and the per-event-type override, default-aware.

745 tests pass locally, clippy/fmt clean. Nice work, this feature landed well. Phase 3 providers can build on the webhook bridge whenever someone needs them.

PR #129, opened 2026-06-05
fix: email verification not blank anymore

arthur-perrot:
Small fix, On public booking pages, entering an incomplete email address (e.g. `arthur@perrot`) passed the native HTML5 validation (`type="email"` accepts domains without a TLD), but then failed server-side, resulting in a blank page displaying only the text: "Please enter a valid email address."

## What changed

- **In-field validation**: added a `pattern` requiring `name@domain.tld` (at least 2 characters after the `.`) along with a localized `title` attribute on the email input in `book.html`. 
 A small script now propagates the message through `setCustomValidity` on `input` and `blur` events, ensuring the browser's native validation tooltip displays the translated text instead of the generic "Please match the requested format" message.
- **Server-side safety net**: all four booking handlers (`handle_booking`, `handle_booking_for_user`, `handle_group_booking`, `handle_dynamic_group_booking`) now route validation failures from `validate_booking_input` through `render_booking_action_error` instead of returning a raw `Html(e)` response.
 This prevents blank pages when JavaScript is disabled or the client-side pattern validation is bypassed.
- Added a new Fluent key: `book-email-invalid` (EN + FR; other locales continue to fall back to English as usual).

Before : 
image

After : 
image

### DIFF
--- a/i18n/en/main.ftl
+++ b/i18n/en/main.ftl
@@ -67,6 +67,7 @@ book-name-label = Your name
 book-name-placeholder = Jane Doe
 book-email-label = Email
 book-email-placeholder = jane@example.com
+book-email-invalid = Please enter a complete email address, including the domain (e.g. jane@example.com).
 book-notes-label = Notes
 book-notes-optional = (optional)
 book-notes-placeholder = Anything you'd like to discuss?
--- a/i18n/fr/main.ftl
+++ b/i18n/fr/main.ftl
@@ -67,6 +67,7 @@ book-name-label = Votre nom
 book-name-placeholder = Jeanne Dupont
 book-email-label = Adresse e-mail
 book-email-placeholder = jeanne@example.com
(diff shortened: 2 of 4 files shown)


olivierlambert (approved):
Good fix, approving. Verified: the client pattern is strictly stricter than `validate_booking_input`, so the server-side path only fires with JS off and now renders a real error page; `setCustomValidity` usage and the `tojson` injection are correct; the four handler call sites match the existing `render_booking_action_error` convention.

One process note for future PRs, no action needed from you this time: **new translatable strings go through the `i18n` branch, not `main`** (see CLAUDE.md, Localization). Hosted Weblate works off the `i18n` branch, so a key added on `main` would never reach the ES/PL translators, and the flow is one-way `i18n -> main`. I've fast-forwarded `i18n` and retargeted this PR's base to it; after merging I'll merge `i18n -> main` so the fix ships immediately and the translations trail in through the normal Weblate cycle.

PR #144, opened 2026-06-23
fix(auth): accept OIDC ID tokens with additional audiences

flokflok1:
## Problem

OIDC login fails against identity providers that put **more than one audience** in the ID token's `aud` claim. The `openidconnect` crate's default verifier (`client.id_token_verifier()`) rejects any audience other than the client_id, so the callback errors out with:

```
Invalid audiences: <some-id> is not a trusted audience
```

This happens with several common IdPs:

- **Zitadel** — always appends the app's `project_id` to `aud` by design (no toggle to disable it), so *every* token has a second audience.
- Some **Keycloak** / **Azure AD** setups that add resource/API audiences.

The result is that cal.rs cannot be used as an OIDC client of these providers at all.

## Fix

Relax the verifier to accept additional audiences via `set_other_audience_verifier_fn`, while still requiring our own client_id to be present.

This is spec-compliant: [OIDC Core §3.1.3.7](https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation) allows the `aud` to contain values beyond the client's own ID as long as `azp` (authorized party) identifies the client — which all the providers above set. Because the verifier independently still requires the client_id to be in `aud`, a token minted for a *different* client is **not** accepted, so this does not weaken audience validation in a meaningful way.

## Testing

- Built against `v1.14.0` and verified end-to-end OIDC login against a Zitadel instance (project_id present in `aud`, `azp` = client_id): login now succeeds where it previously failed with the `Invalid audiences` error.
- Tokens issued for a different client (client_id absent from `aud`) are still rejected.

Happy to make this opt-in behind a config flag instead if you'd prefer to keep strict single-audience checking as the default — let me know.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OIDC sign-in handling during the callback to accept ID tokens that contain additional audience values beyond the app’s configured client ID.
  * Added stricter validation of the token’s authorized party: if an authorized party claim is present, it must exactly match the configured client ID, otherwise the sign-in is rejected.

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -1055,11 +1055,43 @@ async fn oidc_callback(
         None => return Html("No ID token in response.".to_string()).into_response(),
     };
 
-    let verifier = client.id_token_verifier();
+    // Accept ID tokens whose `aud` carries audiences in addition to our
+    // client_id. Several IdPs (e.g. Zitadel, which always adds the project_id;
+    // some Keycloak/Azure AD setups) emit multi-audience tokens. This is
+    // spec-compliant: OIDC Core §3.1.3.7 allows extra audiences. The default
+    // `openidconnect` verifier rejects any audience besides the client_id,
+    // which breaks login against those providers ("Invalid audiences: <x> is
+    // not a trusted audience"). The verifier still independently requires our
+    // own client_id to be present in `aud`, so a token minted for a different
+    // client is still rejected.
+    let verifier = client
+        .id_token_verifier()
+        .set_other_audience_verifier_fn(|_aud| true);
     let claims = match id_token.claims(&verifier, &Nonce::new(stored_nonce)) {
         Ok(c) => c,
         Err(e) => return crate::web::oidc_error_response("oidc id_token verify", &e),
     };
+
+    // Since we now accept additional audiences, restore the `azp` defense that
+    // OIDC Core §3.1.3.7 (5) mandates and that this crate does NOT implement
+    // (its README lists "verification of the azp claim" as unsupported): if the
+    // token carries an `azp` (authorized party) claim, it MUST equal our
+    // client_id. This rejects a multi-audience token that was actually issued
+    // for a *different* client but happens to also list ours. When `azp` is
+    // absent we fall back on the already-verified nonce binding plus the
+    // mandatory client_id-in-aud check, so IdPs that omit `azp` still work.
+    if let Some(azp) = claims.authorized_party() {
+        let azp_is_our_client = auth_config
+            .oidc_client_id
+            .as_deref()
+            .is_some_and(|cid| azp.as_str() == cid);
+        if !azp_is_our_client {
+            return Html(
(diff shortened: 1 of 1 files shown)


flokflok1:
Updated the branch to also re-add the `azp` (authorized party) defense.

This crate's README lists *"verification of the azp claim"* as unsupported, so once we accept additional audiences, a multi-audience token issued for a **different** client (that merely also lists ours in `aud`) would pass. OIDC Core §3.1.3.7 (5) says that when an `azp` claim is present it must equal the client's own ID — the patch now enforces exactly that. When `azp` is absent it falls back on the already-verified nonce binding plus the mandatory client_id-in-`aud` check, so IdPs that don't emit `azp` are unaffected. No regression for single-audience or azp-less providers; the multi-audience case is now validated rather than blanket-trusted.

olivierlambert:
Merged, thank you. The reasoning in this one is better than most auth patches I read.

I checked the load-bearing claim rather than taking it on trust, and it holds: `openidconnect` 4.0.1 really does not verify `azp`. The code is written out in `verification/mod.rs` around the "4. If the ID Token contains multiple audiences" comment and then commented out wholesale, with the maintainers deferring it pending a clear use case. And the verifier does still independently require our own `client_id` in `aud`, so relaxing the *other* audiences does not open the door you'd worry about. Spotting that accepting extra audiences is exactly what promotes `azp` from optional nicety to load-bearing check is the part that makes this patch correct rather than merely working.

I pushed two commits before merging:

- `fix(auth): route the azp rejection through oidc_error_response` — the rejection answered with a bare `Html(...)`, which is HTTP 200, wrote nothing to the log, and named the failing check in the body. Every other verification failure in that handler goes through `oidc_error_response`, which logs at error level and answers 500 with a generic body; there's a test, `oidc_error_response_returns_500_without_detail`, pinning that shape. A security-relevant rejection answering 200 and leaving no trace for the operator was the one thing I wanted changed. The comparison also moved into `azp_matches_client`, a pure function, so the rule is unit-testable without standing up an IdP. Behaviour is unchanged.
- The same commit runs `cargo fmt`, which CI gates on. Worth knowing why you never saw that fail: the CI workflow was parked behind GitHub's first-time-contributor approval gate, so it had never run on this PR at all. The only green check you'd have seen was CodeRabbit. Not your fault, and now fixed.

One thing I left alone deliberately, flagging it so it's on the record rather than lost: a multi-audience token carrying no `azp` is still accepted, where Core section 3.1.3.7 step 4 says one should be present. Your documented reasoning (IdPs that omit `azp` keep working) plus the fact that the ID token arrives over the back-channel code exchange authenticated with our client credentials and PKCE, so it isn't attacker-supplied, makes that a sound trade. If a use case ever argues for tightening it, the check now lives in one testable function.

Verified against current `main`, which moved a fair amount while this sat: merges clean, `cargo fmt --check` clean, clippy clean under `-D warnings`, 860 tests pass. Sorry this waited as long as it did.

PR #149, opened 2026-07-24
feat: shared bookable resources (demo lab, rooms) with availability blocking and CalDAV reservation

### LINKED ISSUE #148: feat: set periods of unavailability for teams events
Today, when an event can not be attended by the whole team for some reasons, the team book themselve their agenda to avoid booking.

This is not convenient. 
We should have the option to designate days when it is not possible to book a slot for team events.

olivierlambert:
Closes #148.

## What this adds

Shared bookable **resources** (a demo lab, a meeting room, a test cluster) as first-class entities backed by a calendar feed. Attach one or more resources to any event type: when a required resource is busy, the event type stops offering slots. Placing a "Lab Maintenance" event in the resource's calendar (from BlueMind or any CalDAV server) now blocks bookings across every event type using that resource, replacing the "every team member books their own agenda" workaround from #148.

## How it works

**Reads (availability) are credential-free.** A resource stores an ICS publish URL (BlueMind "public/private calendar address", Nextcloud public link, any ICS feed). Events are cached in a new `resource_events` table and refreshed on demand when a slot page is viewed (5-minute staleness window, same policy as CalDAV sources). Handles BlueMind's empty-body-with-`text/calendar` response for empty calendars, all-day events with exclusive DTEND, RRULE recurrence, TRANSP and STATUS.

**Two scheduling modes** per event type (`resource_scheduling_mode`):
- `all` (default): every attached resource must be free. Blocking intervals are the union of the resources' busy times.
- `round_robin`: one free resource is enough. Blocking intervals are the intersection (blocked only when every resource is busy), and at booking time the least-loaded free resource is picked and recorded in `bookings.assigned_resource_id`.

**calrs' own bookings block resources too**, independently of feed freshness: a confirmed booking blocks its assigned resource, and in `all` mode every resource of its event type. Two event types sharing a lab cannot double-book it even with no write credentials at all. The check runs inside the booking transaction, alongside the existing member-busy re-check.

**Writes (reservation) are optional.** If a CalDAV collection URL is available (entered manually, or derived automatically from a BlueMind publish URL: `/api/calendars/publish/calendar:UID/...` maps to `/dav/calendars/__uids__/UID/calendar:UID/`), confirmed bookings are PUT into the resource calendar under the booking's own uid and deleted on cancel/decline. Reschedules re-check resource availability, re-pick in round-robin mode, and move the reservation. Credential candidates are tried in order:
1. a service account stored on the resource (AES-256-GCM encrypted, keep-current pattern),
2. members who opted in ("Lend my calendar access for resource reservations" in Profile & Settings) and have a CalDAV source on the same server, with the booking's assigned host preferred.

Write failures are non-fatal and logged: the booking stays confirmed and the internal DB check keeps protecting the resource.

## UI

- **Admin panel**: Resources section with add/edit/delete, per-resource "Sync now" and "Test write" (full PUT/verify/DELETE/verify cycle with a clearly-labeled temporary event), cached-event and attachment counts, and a "writes via" line showing exactly which credentials are used.
- **Event type form** (personal and team): resource checkboxes plus the all/round-robin mode choice.
- **Troubleshoot timeline**: new "Resource busy" segment type so blocked availability is explainable.
- **Settings**: explicit member opt-in for credential lending.

## CLI

`calrs resource probe --url <URL> [--username u] [--write-test]` diagnoses a resource calendar: detects ICS feed vs CalDAV, walks the full RFC 4791 discovery chain when flat PROPFIND returns nothing (BlueMind does), fetches the display name and upcoming events, and optionally proves write access with a temporary event.

## Also in this PR

- `fix(caldav)`: `event_exists()` now falls back to PROPFIND when a server answers HEAD with 501 Not Implemented (BlueMind does; only 405 was handled). This affects existing write-back verification independently of resources and may be worth a cherry-pick to main if this PR takes time to land.
- `fix(resource feed)`: empty `text/calendar` bodies are treated as valid empty calendars.

## Validated against a real BlueMind resource

All of this was probed against a live BlueMind 5 instance with a dedicated "Vates Demo Lab 1" resource:
- publish URL read path (including the empty-calendar quirk and `X-WR-CALNAME` extraction),
- publish endpoints are strictly GET-only ("no service registered on PUT/POST"), so write-back must go through CalDAV,
- authenticated CalDAV PUT and DELETE on the resource collection work with a user holding write ACL (SSO password accepted by DAV Basic auth),
- resource calendars are NOT listed in a member's DAV calendar home, which is why the collection URL is stored/derived explicitly instead of discovered.

## Tests

11 new tests (785 total, all green): both modes in slot computation, `check_and_pick` outcomes, least-loaded round-robin selection, cross-event-type blocking through calrs bookings alone, TRANSPARENT/CANCELLED feed events, all-day exclusive-DTEND conversion, and an end-to-end booking POST asserting `assigned_resource_id`.

## Out of scope (follow-ups)

- Resource capacity > 1 (a resource is booked or free).
- Re-pushing the resource event with updated attendees on watcher claim.
- CLI slot/booking commands do not consult resources yet (web paths do).
- Surfacing the assigned resource in notification emails.

## Summary by CodeRabbit

* **New Features**
  * Added shared bookable resources (e.g., rooms/labs) backed by cached calendar feeds, including resource-based busy-interval blocking.
  * Admin UI now supports full resource management (create, edit, delete, sync) and optional write-back testing.
  * Event types can require resources and choose “all” vs round-robin scheduling; bookings reserve/persist assignments and update resource calendars when enabled.
  * Added `lend_resource_write` opt-in plus a new “Resource busy” availability indicator and troubleshooting updates.
  * Added `resource probe` CLI to inspect feeds and verify calendar write access.
* **Bug Fixes**
  * Improved CalDAV fallback behavior for servers returning HTTP 501.

## Updates since the description was written

- Concurrency: the resource check does NOT run "inside the booking transaction" as originally stated. SQLite reads on separate pool connections cannot see an uncommitted insert, so a process-wide lock now serializes the availability check with the booking insert/update across all four booking flows, the reschedule flow, both approval paths, and `calrs booking create`. Bookings without resources never take the lock, and no network I/O happens while it is held.
- Approval re-check: pending bookings do not block resources, so both approval paths re-verify availability under the lock and re-pick in round-robin mode; if the resource is gone, approval is refused with a visible error instead of confirming a double-booking.
- Admin-only attachment: resources can only be attached to event types by admins. Non-admins never see the form section and their edits leave existing attachments untouched.
- Write scope is mode-aware: `all` mode reserves every attached resource, `round_robin` only the assigned one. Release targets ignore the current mode (it may have been flipped since the reservation was written) and derive from the stored assignment, so detached resources still get released.
- Credential candidates match on scheme + host + port with enabled user, enabled source, and stored password; clearing a resource's service username clears its password with it.
- Feed fetch hardening: no redirects (SSRF), 10 MB body cap, 10 000 cached events per resource, the stored URL re-validated against the private-host policy on every fetch, and failed syncs back off for 5 minutes instead of re-fetching on every request.
- Declined pending bookings need no release: reservations are only pushed on confirmation.
- Test count: 19 new tests, suite total 786 (785 passing, 1 pre-existing ignored).
- Known follow-ups: `calrs event-type slots` does not consult resources yet, and `calrs booking cancel` performs no remote release (it has never done CalDAV cleanup); the cancelled-uid exclusion keeps availability correct either way. The new Fluent keys should be synced to the `i18n` branch after merge.


olivierlambert:
Thanks for the review. Status of each finding, addressed in `d18750a` and `b18ab3e`:

**1. `fetch_feed` follows redirects (SSRF) — fixed.** The client now uses `redirect::Policy::none()`, re-validates the stored URL against the private-host policy on *every* fetch (background re-syncs are triggered from public pages, and the host may start resolving to an internal address after configuration), and bounds the response: 10 MB body cap via chunked read, 10 000 cached events per resource.

**2. Resource check not serialized with the insert (TOCTOU) — fixed, and it was worse than stated.** You are right that the transaction provided no serialization: every read ran on a separate pool connection. Since SQLite cannot give us cross-connection serialization for this pattern, all four booking flows, the reschedule flow, and both approval paths now take a process-wide lock held from the resource check until the booking row is committed, and released before any network write-back. Bonus fixes in the same pass: pending bookings do not block resources, so both approval paths now re-verify (and re-pick in round-robin) instead of confirming a double-booking; failed feed fetches now stamp `last_synced_at` so a dead feed backs off instead of re-fetching with a 30 s timeout on every request.

**3. Route new UI text through Fluent — declining for now, deliberately.** The dashboard/admin surface is intentionally untranslated in this codebase (see the existing dashboard handlers: "stays English until the dashboard surface is translated"); only guest-facing pages use Fluent, and this PR adds no guest-facing text. Localizing just the resources card would make it the only translated island in an English dashboard and put churn on translators before the surface-wide pass. Noted in CLAUDE.md so it gets picked up when the dashboard is localized as a whole.

**4. (nitpick) `same_host` compares hostname only — fixed.** Candidate matching now compares scheme + host + effective port via a shared `url_origin` helper, and the admin "writes via" status line uses the same match, so it no longer claims write-back is enabled when no lender actually matches the resource's origin.

olivierlambert:
Round two, addressed in `87136bc`:

**CLAUDE.md release semantics — fixed.** The write-path section now states explicitly that reservations are only pushed on confirmation, releases happen when a confirmed booking is cancelled or rescheduled, and declined pending bookings need no cleanup.

**Fluent for the new strings — partially accepted, with a correction.** `templates/settings.html` is not "an already-translated template": it contains zero `t()` call sites today, so a lone translated hint there would create exactly the mixed-language island the project convention avoids; it stays English with the rest of that page. However `templates/event_type_form.html` does already use `t()` (the notice fields), so the new "Required resources" section now goes through Fluent keys (`event-type-form-resources-*`) with English entries in `i18n/en/main.ftl`; other languages fall back to English until the next Weblate round on the `i18n` branch.

olivierlambert:
Round three, addressed in `9a85498`: both wording nits were valid. The write-path docs now distinguish the modes (every attached resource in `all`, only the assigned resource in `round_robin`), and the form hint is mode-neutral ("Slots are offered only when the selected resources are available, according to the mode below.").

PR #156, opened 2026-07-25
fix(teams): write-back targets the assigned member; per-member slot uniqueness

### LINKED ISSUE #146: Round-robin: unique index prevents a second booking in the same slot even when other members are free
### Summary

On a team event type in `round_robin` mode, a second booking for the same start
time is rejected with "This slot is no longer available", even when another team
member is free at that time. Only one booking per event type and start time is
possible, regardless of team size.

### Environment

- calrs version: 1.14.0
- Deployment: Docker (`ghcr.io/olivierlambert/calrs:latest`), SQLite
- CalDAV source: SOGo, one source per team member
- 2 team members, both with a working calendar source and equal priority

### Steps to reproduce

1. Create a team with two members, both with a synced CalDAV source
2. Create a team event type, `scheduling_mode = round_robin`, `requires_confirmation` on
3. Book a slot as a guest. The booking is correctly assigned to member A
4. Approve it
5. Book the exact same slot again as a different guest

### Expected

The booking is accepted and assigned to member B, who is free at that time.

### Actual

The booking is rejected with "This slot is no longer available".

Note that the slot picker still offers the slot as available before submitting,
so the availability engine and the write path disagree. The rejection only
happens on submit.

### The constraint is not a capacity limit

Creating a **second** team event type on the **same team** and booking the **same
start time** on it works fine. Both bookings are accepted and round-robin assigns
them to two different members. the availability engine correctly skips the member
already booked.

So calrs is perfectly capable of holding two concurrent bookings for the same team
at the same time; it just refuses them within a single event type. That makes the
restriction look more like an oversight than an intentional capacity model.

### Suspected cause

The partial unique index documented on the security page, and present in the
schema, keys on `event_type_id` and `start_at` only:

```sql
CREATE UNIQUE INDEX idx_bookings_no_overlap
ON bookings(event_type_id, start_at)
WHERE status IN ('confirmed', 'pending');
```

For a round-robin team event type this makes the slot exclusive across the whole
team. Including `assigned_user_id` would presumably preserve the intended
double-booking protection per host while allowing concurrent bookings for
different members.

I have not read the Rust source, so this is inferred from the schema and the
observed behaviour.

Possibly related: #28 (max concurrent bookings per slot) touches the same index,
though it describes a different scenario. several guests in one slot for a single
host, rather than different hosts.

This is arguably the same class of problem as #117 (per-member booking frequency
limits), where the observation was:

> "For a round-robin team event type, `1/day` means the *whole team* takes one
> booking per day, not one per person."

The unique index has the same shape: a constraint that reads as per-host but is
enforced per event type.

### Impact

This makes round-robin unusable for teams that need parallel capacity. Our use
case is a company with several field technicians whose customers book
appointments themselves; with the current behaviour the company could serve
exactly one customer per time slot.

### Happy to help

I have a test environment running against a real SOGo server with three separate
accounts, and the setup is reproducible. If a fix lands, I am glad to test it and
report back.

---

### LINKED ISSUE #147: Team bookings: CalDAV write-back targets the event type owner's calendar, ignoring assigned_user_id
### Summary

On a team event type, confirmed bookings are written to the CalDAV calendar of the
account that **owns the event type**, regardless of which team member the booking
was assigned to. The assigned member never receives the event in their own
calendar.

If the owning account has no CalDAV source at all, write-back is skipped entirely
and silently: no log line, no error, no indication in the UI.

### Environment

- calrs version: 1.14.0
- Deployment: Docker (`ghcr.io/olivierlambert/calrs:latest`), SQLite
- CalDAV sources: SOGo, one per account, each with a write calendar configured
- Team event type owned by a global admin account

### Steps to reproduce

1. Three accounts, each with its own CalDAV source and write calendar. Use three
   **distinct** mailboxes, sharing one between two accounts makes the result
   ambiguous (this produced a false conclusion on our first attempt)
   - admin → `admin@example.com`
   - member A → `user-a@example.com`
   - member B → `user-b@example.com`
2. Admin creates a team containing A and B, and a round-robin team event type
3. Book twice at different times, so round-robin assigns one booking to each member
4. Confirm both from the admin dashboard

### Expected

Each event is written to the calendar of the member in `bookings.assigned_user_id`.

### Actual

Both events are written to the **admin's** calendar, none to the assigned members'.

While `pending`, `caldav_calendar_href` is empty for both bookings, as expected.
On confirmation both are set to the admin's write calendar, although the two
bookings have different `assigned_user_id` values.

Repeating step 4 via the members' own Approve links from the notification emails
makes no difference, so it is not the confirming user that decides the target.

### Also affects collective mode

Switching the same team event type to `collective` produces the same result: the
event lands in the owner's calendar only. The docs describe different behaviour:

> "The booking is created and pushed to **every** member's CalDAV calendar"

So this does not appear specific to round-robin. Write-back for team event types
seems to resolve through the owning account in both scheduling modes.

(Possibly related, same run: the collective booking notified only one member, not
both, where the docs say "Email notifications are sent to all members". I have not
investigated that separately.)

### The intended behaviour appears to be per-member

#31 (sales team notification / demo claim flow) describes the CalDAV event as
living on "the tech's write-back calendar", i.e. the assigned member's. That
matches the docs for collective mode as well. So this looks like a gap rather than
a design decision.

### The documented admin workflow makes this worse

From the Teams page:

> "Global admins can remove themselves from teams they created — they retain
> management access via the admin panel. This supports the IT admin use case of
> creating teams without being bookable."

An admin who is deliberately not bookable has no reason to own a calendar source.
In that configuration nothing is written anywhere, silently. So the documented
setup produces no calendar entries at all, and the configuration that does produce
entries puts them in the wrong calendar.

### Impact

For our use case (several field technicians whose calendars live in SOGo) this
removes the core function: technicians would never see their own appointments, and
every booking for every technician would pile up in one shared calendar.

### Suggestion

Resolve the write target from `bookings.assigned_user_id` for team event types.

This may overlap with #47 (choose write-back calendar per event type), which is
already marked `planned`. that issue is about *which* calendar is selected, this
one about *whose*. They might be worth solving together.

Independently of that: log a warning when write-back is skipped for lack of a
source on the owning account. The docs say skipping is silent when no write
calendar is configured, which is reasonable for a deliberate "None" setting. But
here the members' sources *are* configured, so the silence is misleading. 

### Happy to help

I have a test environment running against a real SOGo server with three separate
accounts, and the setup is reproducible. If a fix lands, I am glad to test it and report back.

---

olivierlambert:
Closes #147, closes #146. Both reported by @3dreams-medienagentur against 1.14.0 with a SOGo test environment.

## #147: write-back resolved through the event type owner

Root cause: `caldav_push_booking(user_id, ...)` found calendars purely by the passed user id, and the confirm paths (dashboard confirm, email-token approve, guest reschedule, claim) all passed the event type owner. The assigned member never got the event; an owner without sources produced silence at debug level.

Fix:
- New `booking_write_targets(booking_uid, fallback)`: the assigned member when `bookings.assigned_user_id` is set, every eligible member for a collective team event type, otherwise the passed fallback (personal host). `caldav_push_booking` and both delete helpers resolve through it, which repairs every call site at once, including CLI cancel.
- The delete side keeps a legacy fallback: if the stored `caldav_calendar_href` matches no write source of the resolved member, it retries with the original owner (bookings pushed before this fix live there), then falls back to delete-by-uid.
- Collective mode was never consulted at booking time, it round-robin-assigned one member with one push and one email. Now: submit-time re-check that every member is free, `assigned_user_id` stays NULL, write-back fans out to every member's calendar, and host notification/approval emails go to every member. Guest-facing emails show the joined member names (same pattern as dynamic groups). This also covers the issue's side observation about only one member being notified.
- The no-sources skip is a `warn!` now, naming the user and booking uid.

## #146: unique index made round-robin slots team-exclusive

Migration 061 rekeys `idx_bookings_no_overlap` to `(event_type_id, start_at, COALESCE(assigned_user_id, ''))`, exactly the shape the reporter inferred. Per-member uniqueness for round-robin; personal, collective (NULL) and dynamic-group bookings keep whole-slot exclusivity.

Two adjacent holes had to close for this to be sound:
- `fetch_busy_times_for_user_ex` marked the event type owner busy for every booking on their event type even when assigned to someone else, and would not have marked members busy for unassigned collective bookings. Attribution is now: assigned member, else owner for unassigned personal bookings, else all team members for unassigned team bookings.
- `pick_group_member` only saw confirmed bookings, but pending ones occupy the index: with the rekeyed index the picker would re-pick a member holding a pending same-slot booking and bounce on INSERT. It now skips members with a confirmed or pending overlapping assignment.

Known limitation (pre-existing, unchanged): two *truly simultaneous* submits can still both pick the same member and the second gets "slot no longer available" — the picker runs outside any lock, and the index remains the final guard. Sequential bookings (the reported case) now work.

## Tests

Five new: per-member index semantics (allowed/rejected/NULL exclusivity), `booking_write_targets` resolution (assigned / collective / personal fallback), busy attribution (owner not busy, members busy for collective), picker pending-slot skip. Full suite: 795 pass, clippy clean.

Docs: booking-flow.md and caldav.md aligned; teams.md already described this behavior and is now accurate.

## Summary by CodeRabbit

* **New Features**
  * Added collective team bookings that schedule across all eligible members and fan out CalDAV write-back to each member calendar.
  * Round-robin bookings now target the assigned member’s calendar.
* **Bug Fixes**
  * Enforced booking uniqueness per assigned member for overlapping pending/confirmed slots.
  * Improved CalDAV orphan cleanup and deletions to respect the correct source/write-calendar boundaries.
  * Refined busy-time and slot-picking behavior for pending bookings.
* **Documentation**
  * Expanded booking-flow and CalDAV write-back docs, including warning behavior when no write calendar is configured.

### DIFF
--- a/docs/src/booking-flow.md
+++ b/docs/src/booking-flow.md
@@ -124,7 +124,7 @@ Additionally, a database-level unique index prevents two bookings from occupying
 
 ## CalDAV write-back
 
-When a booking is confirmed (either directly or via approval), calrs can push the event to the host's CalDAV calendar. For team event types the host is the assigned member (round-robin), or every member (collective). See [CalDAV Integration > Write-back](./caldav.md#caldav-write-back) for setup.
+When a booking is confirmed (either directly or via approval), calrs can push the event to the host's CalDAV calendar. For team event types the host is the assigned member (round-robin), or every eligible member (collective: enabled members with a non-zero per-event-type weight, the same set the slot grid checks). See [CalDAV Integration > Write-back](./caldav.md#caldav-write-back) for setup.
 
 ## Email notifications
 
--- a/docs/src/booking-flow.md
+++ b/docs/src/booking-flow.md
@@ -120,11 +120,11 @@ Before a booking is accepted, calrs checks for conflicts:
 - **Buffer times** — the buffer before/after is included in the conflict window
 - **Minimum notice** — slots too close to the current time are rejected
 
-Additionally, a database-level unique index prevents two bookings from occupying the same slot, even if two guests submit simultaneously.
+Additionally, a database-level unique index prevents two bookings from occupying the same slot, even if two guests submit simultaneously. On round-robin team event types the uniqueness is per assigned member, so two guests can book the same time as long as different members take the bookings.
 
 ## CalDAV write-back
 
-When a booking is confirmed (either directly or via approval), calrs can push the event to the host's CalDAV calendar. See [CalDAV Integration > Write-back](./caldav.md#caldav-write-back) for setup.
+When a booking is confirmed (either directly or via approval), calrs can push the event to the host's CalDAV calendar. For team event types the host is the assigned member (round-robin), or every member (collective). See [CalDAV Integration > Write-back](./caldav.md#caldav-write-back) for setup.
 
 ## Email notifications
 
--- a/docs/src/caldav.md
+++ b/docs/src/caldav.md
@@ -93,8 +93,8 @@ When a booking is confirmed, calrs can automatically push it to your CalDAV cale
 - On **confirmation**: calrs generates an ICS event and PUTs it to `{calendar-href}/{booking-uid}.ics`
 - On **cancellation**: calrs DELETEs the event from the same path
 - The booking tracks which calendar it was pushed to, so cancellation always targets the right calendar
-- If no write calendar is configured, write-back is silently skipped (emails still work)
-- Write-back works for individual bookings, group round-robin bookings, and pending-then-confirmed bookings
+- If no write calendar is configured, write-back is skipped and a warning is logged (emails still work)
+- Write-back works for individual bookings, team bookings (the assigned member's calendar for round-robin, every member's for collective), and pending-then-confirmed bookings
 
 ## Managing sources
 
--- a/docs/src/caldav.md
+++ b/docs/src/caldav.md
@@ -94,7 +94,7 @@ When a booking is confirmed, calrs can automatically push it to your CalDAV cale
 - On **cancellation**: calrs DELETEs the event from the same path
 - The booking tracks which calendar it was pushed to, so cancellation always targets the right calendar
 - If no write calendar is configured, write-back is skipped and a warning is logged (emails still work)
-- Write-back works for individual bookings, team bookings (the assigned member's calendar for round-robin, every member's for collective), and pending-then-confirmed bookings
+- Write-back works for individual bookings, team bookings (the assigned member's calendar for round-robin, every eligible member's for collective), and pending-then-confirmed bookings
 
 ## Managing sources
 
--- a/migrations/061_booking_unique_per_member.sql
+++ b/migrations/061_booking_unique_per_member.sql
@@ -1,0 +1,8 @@
+-- Round-robin team event types: one slot must be bookable once per assigned
+-- member, not once per event type (#146). Rekey the double-booking guard by
+-- the assigned member; NULL (personal event types, collective teams, dynamic
+-- groups) keeps the old whole-slot exclusivity.
+DROP INDEX IF EXISTS idx_bookings_no_overlap;
+CREATE UNIQUE INDEX idx_bookings_no_overlap
+ON bookings(event_type_id, start_at, COALESCE(assigned_user_id, ''))
+WHERE status IN ('confirmed', 'pending');
--- a/src/commands/sync.rs
+++ b/src/commands/sync.rs
@@ -1022,6 +1022,7 @@ async fn cancel_orphaned_booking(
          JOIN accounts a ON a.id = et.account_id
          JOIN users u ON u.id = COALESCE(b.assigned_user_id, a.user_id)
          JOIN caldav_sources cs ON cs.id = ?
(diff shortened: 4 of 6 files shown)


olivierlambert:
@coderabbitai review

olivierlambert:
@coderabbitai full review

olivierlambert:
@coderabbitai full review

olivierlambert:
CodeRabbit triage (5fba015):

- Fixed: orphan sweep now also requires the synced source to belong to the booking's effective host (assigned member, else owner) — hrefs are not globally unique across servers (Major, confirmed).
- Fixed: guest-reschedule pending path uses caldav_delete_booking, keeping the pre-#147 owner-calendar fallback before the href pointer is cleared.
- Fixed: picker comment now states the broad exclusion (any event type, buffered window) is intentional; a member with an overlapping pending commitment anywhere is a bad pick.
- Fixed: docs say "every eligible member" for collective write-back.
- Declined: scoping the orphan uid lookup to the source's own cached calendars. Confirm-before-cancel is not available on every sweep path, and a write calendar absent from the synced set would then read as "everything orphaned" and false-cancel. The global lookup errs on the lenient side by design; comment added in the code.

PR #160, opened 2026-07-28
fix(teams): keep the booking's organizer when rescheduling

### LINKED ISSUE #159: Rescheduling a team booking silently fails to update the calendar (Google CalDAV `403 Forbidden`)
## Summary

When a booking on a **team event type** is rescheduled, calrs updates its own database but the
change never reaches the connected calendar. The guest is shown *"Your booking has been
rescheduled."*, the dashboard shows the new time — and the event on the host's Google Calendar
stays at the **old** time, forever.

Nothing surfaces the failure. The calendar and calrs simply disagree from that point on, and
whoever relies on the calendar shows up at the wrong time.

Bookings on personal (non-team) event types are not affected.

## Environment

| | |
|---|---|
| calrs | 1.15.0 (`ghcr.io/olivierlambert/calrs:latest`) |
| Calendar backend | Google Calendar over CalDAV, connected via OAuth2 |
| Event type | team, round-robin, 2 members, no confirmation required |
| Deployment | Docker Compose |

## Steps to reproduce

1. Create a team event type with two members, round-robin, no confirmation required.
2. Connect a Google Calendar for each member and select a write calendar for each.
3. Book a slot as a guest. Round-robin assigns it to one member, and the event correctly
   appears on that member's Google Calendar. ✅
4. Open the reschedule link from the confirmation email and pick another slot.
5. calrs confirms the reschedule. Check the member's Google Calendar — **the event is still at
   the original time.**

## Expected vs actual

| | |
|---|---|
| Expected | the event moves to the new time on the member's calendar |
| Actual | the event stays at the old time; calrs shows the new one |

In my case, after rescheduling from 14:30 to 16:00 (Europe/Paris):

| | time |
|---|---|
| calrs booking | 16:00 |
| event on Google, same UID | 14:30 — unchanged |

## What the log shows

At `RUST_LOG=calrs=debug`:

```
DEBUG calrs::web: pushing booking to calendar
  uid=<uid>@calrs calendar_href=/caldav/v2/member%40example.com/events/ provider=caldav

ERROR calrs::web: calendar write-back failed
  uid=<uid>@calrs
  error=PUT https://apidata.googleusercontent.com/caldav/v2/member%40example.com/events/<uid>@calrs.ics
        returned 403 Forbidden — <?xml version="1.0" encoding="UTF-8"?>
        <D:error xmlns:D="DAV:"/>
```

Google gives no reason — just an empty `<D:error/>`.

Worth noting on its own: this only ever reaches the log. The reschedule is reported as
successful to the guest and to the host, and nothing in the dashboard indicates that the
calendar was not updated. A reschedule that failed visibly would have been noticed the same
day; this one went unnoticed until the calendars were compared by hand.

## Cause

The two ICS files disagree on who organizes the event.

The event that the **booking** flow wrote to the calendar names the **assigned member** as
organizer:

```
ORGANIZER;CN=member@example.com:mailto:member@example.com
```

The ICS that the **reschedule** flow pushes names the **event type owner** instead:

```
ORGANIZER;CN=Owner Name:mailto:owner@example.com
```

Google CalDAV lets you set any organizer when an event is *created*, but refuses to *change*
the organizer of an event that already exists — so it rejects the update with `403`.

In short: the reschedule flow resolves the host from the event type's owner rather than from
the booking's assigned member. It writes to the correct calendar, but with the wrong identity.

## Control test — it is not "Google refuses updates"

To rule out Google rejecting CalDAV updates in general, I repeated the test on a **personal**
event type, where the organizer is the same at booking time and at reschedule time — same
Google account, same calendar, same credentials:

- 1st reschedule → event created on the calendar ✅
- 2nd reschedule → existing event **updated** ✅

Updates work fine. What Google rejects is the organizer changing.

## Scope

Two other flows stamp the organizer the same way — approving a pending booking from the
dashboard, and approving it through the emailed link. Neither fails today, because both run
right after the event was deleted and re-created (and a *create* may set any organizer), but
both pin the wrong organizer. So fixing only the reschedule path would just move the failure to
the next reschedule; all three need to agree.

## Fix

I have a patch that resolves the host identity from the booking — assigned member for
round-robin, the full roster for collective, the owner for personal event types — and applies it
to all three flows, plus a regression test covering the three cases.

Verified:

- full test suite passes (797 tests)
- against a real Google Calendar, the same reschedule goes from `403 Forbidden` to
  `calendar write-back succeeded`, and the event's start time follows the new slot

Happy to open a PR with the technical detail.

It may also be worth surfacing a failed calendar write-back somewhere in the UI, independently
of this fix.

## Related, but separate

While tracing this I noticed the reschedule flow also checks availability against the event
type owner rather than the assigned member. In my case that let a guest reschedule onto a slot
that was already busy on the assigned member's calendar, and the slot grid offered to the guest
is computed for the wrong person too. I can file that separately — the fix touches the
reschedule page itself, so it does not belong in the same change.

hugo-fasone:
Fixes #159.

## Problem

Rescheduling a booking on a team event type updates `bookings.start_at` but the CalDAV `PUT` is
rejected by Google with `403 Forbidden`. The error is only logged, so the guest and the host are
both told the reschedule succeeded while the calendar keeps the old time indefinitely.

The reschedule handler resolves the host through the event type's account owner and ignores the
booking's `assigned_user_id` (`src/web/mod.rs:18011`):

```sql
FROM bookings b
JOIN event_types et ON et.id = b.event_type_id
JOIN accounts a ON a.id = et.account_id
JOIN users u ON u.id = a.user_id          -- owner, not the assigned member
```

`host_email` is then read from that user (`src/web/mod.rs:18241`), and both fields end up in the
ICS as `ORGANIZER;CN={host_name}:mailto:{host_email}` (`src/email.rs:518`, `src/email.rs:575`).

The booking flow, meanwhile, uses the assigned member: `pick_group_member` returns
`(member_id, name, email)` (`src/web/mod.rs:9850`). So the event is created with the member as
organizer and re-pushed with the owner as organizer. Google CalDAV accepts an arbitrary
`ORGANIZER` on create but refuses a `PUT` that *changes* an existing event's organizer.

Observed on the same booking, same calendar:

```
# event on Google, written by the booking flow
ORGANIZER;CN=member@example.com:mailto:member@example.com

# ICS pushed by the reschedule flow
ORGANIZER;CN=Owner Name:mailto:owner@example.com
```

Note the write-back **target** is already correct — `booking_write_targets`
(`src/web/mod.rs:18644`) prefers `assigned_user_id` since dbd1a64. This PR fixes the identity
written into it, so the pair is consistent.

## Change

New `booking_host_identity(pool, booking_uid) -> Option<(String, String)>`, placed next to
`booking_write_targets` and resolving the same way the booking flow does:

| event type | organizer |
|---|---|
| round-robin / assigned | the assigned member |
| collective | members joined with ` & `, first member's address |
| personal | the account owner |

The collective branch reuses the booking flow's eligibility filter
(`u.enabled = 1 AND COALESCE(etw.weight, 1) > 0`) and its `ORDER BY u.name`, otherwise the
organizer drifts again as soon as a member is added.

`accounts.user_id` is nullable (`ON DELETE SET NULL`, and team event types hang off a placeholder
account), so it is typed `Option` and the helper returns `None` rather than assuming an owner
exists. Each call site keeps its previous behaviour on `None`.

### Applied to three call sites, not one

Every flow that pushes an ICS has to agree on the organizer, or the mismatch just moves:

| call site | was |
|---|---|
| `guest_reschedule_booking` (`src/web/mod.rs:18241`) | event type owner — the reported bug |
| `approve_booking_by_token` (`src/web/mod.rs:17020`) | event type owner |
| `confirm_booking` (`src/web/mod.rs:4400`) | whoever clicked approve |

The two approval paths do not `403` today, because a guest reschedule on a `requires_confirmation`
event type deletes the CalDAV event and re-pushes it on approval — and a *create* may set any
organizer. But they pin the wrong organizer, so fixing only the reschedule path would have made
the *next* reschedule fail instead.

## Testing

`booking_host_identity_follows_the_booking_not_the_owner`, modelled on the existing
`booking_write_targets_resolves_assigned_collective_and_fallback`, covers assigned / collective /
personal.

Local run of the CI gates on this exact commit, with `RUSTFLAGS="-D warnings"`:

- `cargo fmt --check` — clean
- `cargo clippy --profile=test` — clean
- `cargo test` — 797 passed, 0 failed, 1 ignored

### Verified against a real Google Calendar

Same booking, same account, same OAuth2 source, before and after the patch:

| | before | after |
|---|---|---|
| log | `calendar write-back failed … 403 Forbidden` | `calendar write-back succeeded` |
| event `DTSTART` | frozen at the pre-reschedule value | follows the new slot |

`LAST-MODIFIED` on the re-synced event matches the write-back timestamp, confirming Google
accepted the write rather than the change being a local artefact.

**Control test** — to rule out Google rejecting CalDAV updates in general, the same sequence on a
**personal** event type (organizer identical at create and at reschedule), same calendar and
credentials: first push creates the event, second push **updates** it, both succeed. Google is not
refusing updates; it is refusing the organizer change.

## Notes for review

- Touches the Google Calendar path, so @bboles may want to weigh in per `MAINTAINERS.md`. The
  change itself is provider-agnostic — it only changes which user the organizer is taken from.
  Google is the only backend I have to test against, so the `403` above is the only server
  behaviour I can report first-hand.
- **Out of scope, happy to file separately:** the same handler checks availability against the
  owner rather than the assigned member (`src/web/mod.rs:18110`), and `guest_reschedule_slots` has
  the same issue — so the slot grid offered to the guest is computed for the wrong person. In my
  case this let a guest reschedule onto a slot already busy on the assigned member's calendar.
  Fixing it touches the reschedule page (the host profile and avatar come from the same query), so
  it does not belong in this change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Calendar invitations now identify the correct booking host based on the actual booking organizer details.
  * Improved host details for approvals, reschedules, and collective/unassigned bookings.
  * Added smarter fallback behavior when a booking has no explicit host assigned.

### DIFF
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -4389,6 +4389,22 @@ async fn confirm_booking(
     let (date, start_time, end_time) =
         booking_strings_in_guest_tz(&start_at, &end_at, stored_tz, guest_tz_parsed);
 
+    // The organizer belongs to the booking, not to whoever clicked approve:
+    // on a team event type the approver is often the owner while the event
+    // lives on the assigned member's calendar. Writing the approver here would
+    // pin the wrong ORGANIZER and make every later reschedule 403 against
+    // Google CalDAV. See `booking_host_identity`.
+    let (host_name, host_email) = booking_host_identity(&state.pool, &uid)
+        .await
+        .unwrap_or_else(|| {
+            (
+                user.name.clone(),
+                user.booking_email
+                    .clone()
+                    .unwrap_or_else(|| user.email.clone()),
+            )
+        });
+
     let details = crate::email::BookingDetails {
         event_title,
         date,
@@ -4397,11 +4413,8 @@ async fn confirm_booking(
         guest_name,
         guest_email,
         guest_timezone,
-        host_name: user.name.clone(),
-        host_email: user
-            .booking_email
-            .clone()
-            .unwrap_or_else(|| user.email.clone()),
+        host_name,
+        host_email,
         uid: uid.clone(),
         notes: None,
         location: location_display,
@@ -17016,13 +17029,22 @@ async fn approve_booking_by_token(
         booking_strings_in_guest_tz(&start_at, &end_at, stored_tz, guest_tz_parsed);
     let date_label = format_date_label(&date, lang);
 
-    // Get host email for BookingDetails
-    let host_email: String =
-        sqlx::query_scalar("SELECT COALESCE(booking_email, email) FROM users WHERE id = ?")
-            .bind(&user_id)
-            .fetch_one(&state.pool)
-            .await
-            .unwrap_or_default();
+    // Host side for BookingDetails: resolved from the booking so the ORGANIZER
+    // matches the calendar the event is written to. Approving through the email
+    // token used to stamp the event type owner even for a booking assigned to
+    // another member, which later made Google CalDAV 403 the reschedule PUT.
+    // See `booking_host_identity`.
+    let (host_name, host_email) = match booking_host_identity(&state.pool, &uid).await {
+        Some(identity) => identity,
+        None => (
+            host_name,
+            sqlx::query_scalar("SELECT COALESCE(booking_email, email) FROM users WHERE id = ?")
+                .bind(&user_id)
+                .fetch_one(&state.pool)
+                .await
+                .unwrap_or_default(),
+        ),
+    };
 
     // Status is now 'confirmed' (we just transitioned it). Generate the
     // auto meeting URL if the event type uses one; otherwise fall back to
@@ -18238,12 +18260,21 @@ async fn guest_reschedule_booking(
         tracing::info!(booking_id = %booking_id, old_start = %old_start_at, new_start = %new_start_at, "booking rescheduled by guest (auto-confirmed)");
     }
 
-    let host_email: String =
-        sqlx::query_scalar("SELECT COALESCE(booking_email, email) FROM users WHERE id = ?")
-            .bind(&host_user_id)
-            .fetch_one(&state.pool)
-            .await
-            .unwrap_or_default();
+    // Resolve the host side from the booking, not from the event type owner:
+    // the ICS pushed below has to carry the same ORGANIZER the event was
+    // created with, or Google CalDAV rejects the PUT with 403 Forbidden and the
+    // reschedule never reaches the calendar. See `booking_host_identity`.
+    let (host_name, host_email) = match booking_host_identity(&state.pool, &uid).await {
+        Some(identity) => identity,
+        None => (
+            host_name,
+            sqlx::query_scalar("SELECT COALESCE(booking_email, email) FROM users WHERE id = ?")
+                .bind(&host_user_id)
+                .fetch_one(&state.pool)
+                .await
+                .unwrap_or_default(),
+        ),
+    };
 
     // old_start_at/old_end_at are stored in the event-type tz. Convert into the
     // guest's tz so `RescheduleDetails` carries guest-local wall-clock for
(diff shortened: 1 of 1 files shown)


hugo-fasone:
Thanks for the review. Taking the three findings in turn.

**Empty organizer email in the `None` fallback (both call sites)** — this pattern is unchanged from the base commit. At `318c62d`, both call sites already read:

```rust
let host_email: String =
    sqlx::query_scalar("SELECT COALESCE(booking_email, email) FROM users WHERE id = ?")
        .bind(&user_id)
        .fetch_one(&state.pool)
        .await
        .unwrap_or_default();
```

This PR moves it into a fallback arm that is now reached far less often: in both handlers the booking has just been selected through an inner join on `users u ON u.id = a.user_id`, so `booking_host_identity` returns `Some` unless the row disappears between two queries. I would rather not change that error semantics here — deciding what a booking with no resolvable host should do is a separate call.

The "no log trail" part is fair though, so 9872365 adds a `tracing::warn!` to both fallback arms. Silently writing a malformed `ORGANIZER` is precisely what made #159 hard to trace, so that seems worth having regardless.

**Collective organizer drift** — this one is a real gap, now filed as #161. `booking_host_identity` rebuilds the roster from *current* eligible members, so if the team composition changes after the event was written, `first_email` can point at someone else and the PUT gets the same 403 back. I noted the drift risk in the code comment but only guarded against query divergence, not against membership itself changing.

Two corrections to the finding:

- Assigned (round-robin) bookings do *not* drift. The organizer is resolved by user id, which is stable. The risk is specific to collective bookings, where `assigned_user_id` is NULL by design.
- Collective was already broken before this PR, and more broadly: creation used the first member's address while reschedule used the owner's, so the two disagreed whenever the owner was not first by name. This change narrows the failure from "always" to "only if the roster changes".

Persisting the organizer on the booking is the right fix, but `bookings` has no column for it, so it needs a migration plus a backfill decision for existing rows. That is more than this PR should carry, hence the separate issue — happy to take it on if the approach in #161 sounds right.

olivierlambert (approved):
Reviewed and verified locally: the helper mirrors the booking flow queries exactly (same eligibility filter and ordering), fmt clean, 797 tests passing on the PR head. Excellent write-up and root-cause analysis. Please do file the follow-up issue about reschedule availability being checked against the owner instead of the assigned member.

olivierlambert:
Your fix just shipped in [v1.15.1](https://github.com/olivierlambert/calrs/releases/tag/v1.15.1). Thanks a lot for the contribution, the root-cause analysis and the live verification against Google Calendar made this a pleasure to review.

Could you upgrade and give it a spin as soon as you get a chance? In particular the flows this touched: guest reschedule on a round-robin booking, approval via the email token, and approval from the dashboard. I am counting on your field feedback to confirm nothing regressed functionally, since Google is a backend you can exercise for real and I cannot.

I also filed the out-of-scope bug you flagged (reschedule availability computed against the owner instead of the assigned member) as #166 so it does not get lost. If you feel like taking that one too, it is all yours.

hugo-fasone:
Great! Will try to tackle it ASAP

hugo-fasone:
Tested the latest version, no regressions from my end from this PR

PR #178, opened 2026-08-10
fix(theme): derive a real dark palette for custom themes

### LINKED ISSUE #177: Dark mode appears to be broken
When toggling the application to dark mode, only some UI elements seem to change (mainly borders), while the rest of the interface remains white/light. The dark theme colors are not being applied to the page backgrounds, text, inputs, cards, and other UI components.

Image
Image

olivierlambert:
Fixes #177.

## Root cause

`custom_theme_css()` emitted the configured colors verbatim in **both** the `:root` and the `html.dark` block:

```rust
":root{--bg:{bg};--surface:{surface};--text:{text};…}\
 html.dark{--bg:{bg};--surface:{surface};--text:{text};…}"
```

Only the alpha of the derived `--accent-subtle` / `--accent-border` differed between the two. So under a custom theme, toggling dark mode kept the light background, surface and text.

That also explains the exact symptom in the report ("only the borders change"): the block never sets `--border`, `--border-hover`, `--surface-hover`, `--text-secondary` or `--text-muted`, so those *do* fall through to `base.html`'s `html.dark` palette and turn dark, while everything the block pins stays white. Dark borders on a white page.

Preset themes were never affected: `preset_theme_css()` ships a hand-picked dark palette per preset.

## Fix (option A from the issue discussion)

The admin panel collects one set of five colors, which are taken to be the light palette. The dark counterpart is now derived from them:

- **Hue and a clamped saturation are kept**, so a brand tint survives into dark mode. Lightness is re-mapped onto the scale `base.html` already uses (bg `0.04`, surface `0.095`, text `0.97`).
- The neutral ramp is anchored on the **background** hue, not the surface: a surface of `#ffffff` carries no hue of its own, so deriving from it would drop the tint.
- **Supporting neutrals** (`--surface-hover`, `--border`, `--border-hover`, `--text-secondary`, `--text-muted`) are offsets from the resolved surface and text, so they stay in the same color family, and the dark block now sets the full variable set the way the presets do.
- A palette that is **already dark** (background lightness below 50%) is passed through as configured; only the supporting neutrals get derived, so its borders match its surface.
- Saturation is clamped (`0.40` for neutrals, `0.15` for text) and secondary/muted text has a lightness floor, so an unusual palette cannot produce something unreadable.

Two small helpers were added next to `custom_theme_css`: `hex_to_hsl()` and `hsl_to_hex()`, both reusing the existing `hex_to_rgb()`.

## Sanity check

Feeding the *default* light colors through the derivation reproduces the built-in dark palette to within a shade, which is a good sign the lightness targets are right:

| Variable | Derived | `base.html` dark |
|---|---|---|
| `--bg` | `#0a0a0b` | `#09090b` |
| `--surface` | `#171719` | `#18181b` |
| `--border` | `#242428` | `#27272a` |
| `--text` | `#f7f7f8` | `#fafafa` |
| `--text-secondary` | `#a0a0ac` | `#a1a1aa` |

A warm off-white brand palette (`#f5f4f0` / `#ffffff` / `#1a1b38`) yields `--bg:#0c0b08`, `--surface:#1d1b13`, `--border:#2e2b1e`, `--text:#f6f6f8`: dark, readable, and still recognisably warm.

## Tests

Four unit tests: HSL round-trip, the light-palette derivation (regression test for this issue), pass-through of an already-dark palette, and rejection of invalid hex. Full suite green (801 passed).

## Notes / possible follow-ups

- **Light mode is untouched.** Its supporting neutrals still come from `base.html`'s defaults, so no existing custom theme changes appearance in light mode. Only the `html.dark` block is new.
- **The accent is left exactly as configured in both modes**, as agreed. A very dark brand accent (say a navy) will have poor contrast on the derived near-black background. Worth a separate look if anyone hits it; rescuing it means shifting the brand color, which felt out of scope here.
- If someone wants pixel control later, explicit `custom_bg_dark` / `custom_surface_dark` / `custom_text_dark` columns could be added, prefilled with these derived values.

## Summary by CodeRabbit

* **New Features**
  * Added automatic dark-theme color derivation for light and mid-tone palettes.
  * Preserved existing dark palettes and accent colors.
  * Improved support for custom themes, including stored and default configurations.

* **Bug Fixes**
  * Improved handling of whitespace and invalid color inputs to prevent theme generation errors.

* **Tests**
  * Added coverage for color conversion, palette derivation and preservation, custom themes, whitespace handling, and invalid inputs.

### DIFF
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -13275,6 +13275,12 @@ fn custom_theme_css(
     {
         return String::new();
     }
+    // validate_hex trims before measuring, and the admin form stores whatever
+    // was pasted, so trim here too: byte-slicing an untrimmed accent below
+    // would read the wrong nibbles and silently emit rgba(0,0,0,..).
+    let (accent, accent_hover) = (accent.trim(), accent_hover.trim());
+    let (bg, surface, text) = (bg.trim(), surface.trim(), text.trim());
+
     // Parse accent for subtle/border/muted derivations
     let r = u8::from_str_radix(&accent[1..3], 16).unwrap_or(0);
     let g = u8::from_str_radix(&accent[3..5], 16).unwrap_or(0);
@@ -13283,9 +13289,13 @@ fn custom_theme_css(
     // Anchor the dark neutrals on the background hue: a surface of #fff carries
     // no hue of its own, so deriving it from the surface would drop the tint.
     let (bg_h, bg_s, bg_l) = hex_to_hsl(bg).unwrap_or((0.0, 0.0, 1.0));
-    let (text_h, text_s, _) = hex_to_hsl(text).unwrap_or((0.0, 0.0, 0.0));
+    let (text_h, text_s, text_l) = hex_to_hsl(text).unwrap_or((0.0, 0.0, 0.0));
     let neutral_s = bg_s.min(DARK_NEUTRAL_MAX_SAT);
-    let (dark_bg, dark_surface, dark_text) = if bg_l >= 0.5 {
+    // Text darker than its background means the palette was designed for light
+    // mode, whatever its absolute lightness. Testing the pair rather than the
+    // background alone also covers mid-tone backgrounds, which an absolute
+    // threshold would leave with dark mode still looking like light mode.
+    let (dark_bg, dark_surface, dark_text) = if text_l < bg_l {
         (
             hsl_to_hex(bg_h, neutral_s, DARK_BG_L),
             hsl_to_hex(bg_h, neutral_s, DARK_SURFACE_L),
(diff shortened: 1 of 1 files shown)


olivierlambert:
## Red-team pass on this PR

Adversarial review of my own change. Two findings worth fixing, pushed as 6926600; the rest are recorded here as cleared or as known limits.

### Fixed: mid-tone backgrounds still had the original bug

Classifying the palette by background lightness alone (`bg_l >= 0.5`) left anything just under the threshold on the pass-through path, where the dark block repeats the light colors. That is issue #177 all over again for a mid-tone palette.

Comparing the pair is a truer test: text darker than its background means the palette was drawn for light mode, whatever its absolute lightness. Any usable light palette has dark text on a lighter background, and any usable dark palette has the reverse, so both classify correctly, and mid-tones now derive properly. It is also better in the other direction: light text on a mid-grey background (lightness above 0.5) was previously flipped to near-black against the admin's intent, and is now passed through.

### Fixed: whitespace in a stored color silently corrupted the accent

`validate_hex` trims before measuring the length, but the byte-slicing that derives the rgba accent variants ran on the untrimmed value, so slot 1..3 read `"#b"` instead of `"be"`. Verified empirically by disabling the fix:

```
" #be1621 "  ->  --accent-subtle:rgba(0,225,98,0.08)
```

A green tint derived from a red accent, silently. Reachability is limited: the admin form's named inputs are `<input type="color">`, which always submit a normalized 7-char value, so this needs a hand-crafted POST to an admin-only endpoint. Hardening rather than a live user-facing bug, but the function should not depend on its caller for that. Both findings pre-date this branch in spirit; the whitespace one is untouched original code.

### Checked and cleared

- **Cascade.** `/accent.css` is linked at `base.html:442`, immediately after the base `<style>` closes at 441. Equal specificity, later wins. The only later `html.dark` rules are the embed brand block (accent variables only) and `.badge-warning` / theme-pill rules, none of which touch bg, surface or text.
- **Hard-coded light colors.** Since this PR is what makes dark mode actually engage for these users, any pre-existing dark-mode gap in the templates would surface now. `dashboard_base.html` has exactly two literal colors: white text on the accent-filled avatar circle, and the orange in the wordmark. Both are correct on a dark background.
- **Calibration.** Feeding the default light palette through the derivation reproduces the built-in dark palette to within a shade on every variable, so the lightness targets are not arbitrary.
- **Saturation blowout.** Chroma collapses near the ends of the lightness scale, so even a fully saturated light tint cannot produce a garish dark surface: `#ffe8e8` yields `#0e0606` / `#220f0f`. The clamps matter only for `--border-hover` in the middle of the scale.
- **Contrast.** Derived secondary and muted text land near lightness 0.63 to 0.65 against a 0.04 background, comfortably past 4.5:1, with a floor so an unusual palette cannot sink them into the background.
- **Numerics.** The saturation denominator cannot divide by zero, since lightness 0 or 1 implies a zero delta, which returns early. Hue is `rem_euclid`-normalized on both directions of the conversion, and the sextant index cannot fall out of range. Round-trip test over six colors including pure white and pure black.
- **Empty-string columns.** `theme='custom'` with an empty string (not NULL) in a color column fails validation and blanks the whole theme, silently falling back to default blue. Not reachable through the form, and unchanged by this PR, so left alone. Flagging it since `org_accent_hex` already guards against empties, which suggests it has been a concern somewhere.

### Known limits, deliberately out of scope

- **Light mode for a dark-first palette** keeps the base light borders against the admin's dark surface, the mirror image of the bug fixed here. Pre-existing, and a full fix means deriving in both directions, which is a bigger change than this PR should carry.
- **Dark-first palettes now get different dark-mode borders** than before, derived from their own surface rather than the base `#27272a`. That is a visible change for those deployments, and an improvement: the border used to be darker than the surface it outlined. Covered by a test.
- **Accent contrast** in dark mode is still whatever the admin configured, as agreed.
- **No page was rendered.** Everything here is verified at the CSS-generation level plus the DB read; the visual pass on a real instance is still worth doing before merge.

PR #180, opened 2026-08-12
feat: add SMS notifications via Twilio

Chr1s16:
<html>
<body>
<html><head></head><body><h1>SMS Notifications via Twilio</h1><p>Adds opt-in SMS booking notifications through the Twilio REST API, following the existing SMTP pattern throughout the codebase.</p><h2>What this does</h2><ul><li><p><strong>Per-event-type opt-in</strong> — Adds an <strong>SMS notifications</strong> checkbox to the event type form, enabling SMS notifications only for that event type. Existing event types default to disabled, so there is no behaviour change for existing installations.</p></li><li><p><strong>Guest phone number</strong> — When SMS notifications are enabled for an event type, the booking form displays an optional phone number field. The guest's number is normalised to E.164 format on the server.</p></li><li><p><strong>SMS notifications</strong> — Sends notifications for:</p><ul><li><p>Booking confirmation (or after approval for pending bookings)</p></li><li><p>Host-initiated cancellation</p></li><li><p>Guest-initiated cancellation</p></li></ul></li><li><p><strong>Admin panel</strong> — Adds an <strong>SMS settings (Twilio)</strong> section where administrators can configure:</p><ul><li><p>Twilio Account SID</p></li><li><p>Twilio Auth Token</p></li><li><p>From number</p></li><li><p>Default country code for normalising locally formatted guest numbers</p></li></ul></li><li><p><strong>Environment variable override</strong> — <code inline="">CALRS_TWILIO_ACCOUNT_SID</code>, <code inline="">CALRS_TWILIO_AUTH_TOKEN</code>, and <code inline="">CALRS_TWILIO_FROM_NUMBER</code> override the database configuration, following the same <strong>"full block wins"</strong> semantics as <code inline="">CALRS_SMTP_*</code>. The admin form is locked while the environment-variable configuration is active.</p></li><li><p><strong>Dashboard bookings</strong> — Displays the guest's phone number inline next to their email address when SMS is enabled for that booking.</p></li></ul><h2>Files changed</h2>
File | Change
-- | --
src/sms.rs | New module — Twilio REST client, phone number normalisation, and country code list
migrations/062_sms_notifications.sql | Adds phone_number to bookings, sms_notifications_enabled to event_types, and the twilio_config table
migrations/063_sms_default_country_code.sql | Adds default_country_code to twilio_config
src/db.rs | Registers migrations 062 and 063
src/main.rs | Adds mod sms
src/models.rs | Adds sms_notifications_enabled to EventType and phone_number to Booking
src/web/mod.rs | Adds Twilio admin routes and extends booking/cancellation/approval flows with SMS
templates/admin.html | Adds Twilio configuration UI
templates/book.html | Adds optional phone number field, shown only when SMS is enabled
templates/dashboard_bookings.html | Displays the guest phone number next to their email
templates/event_type_form.html | Adds the SMS notifications checkbox

<h2>Design decisions</h2><h3>No new Cargo dependencies</h3><p>Twilio's REST API is plain HTTPS, so the implementation reuses the existing <code inline="">reqwest</code> client already used for CalDAV instead of adding a Twilio SDK.</p><h3>Purely opt-in, zero default behaviour change</h3><p><code inline="">sms_notifications_enabled</code> defaults to <code inline="">0</code>, and <code inline="">phone_number</code> is nullable. Installations with no Twilio configuration and no event types configured for SMS behave exactly as before.</p><h3>Auth token encrypted at rest</h3><p>The Twilio auth token uses the existing <code inline="">crate::crypto</code> implementation (AES-256-GCM), the same mechanism used for SMTP passwords.</p><h3>Pending bookings</h3><p>SMS notifications are deferred until a pending booking is approved, matching the existing email notification behaviour.</p><h3>Phone number normalisation</h3><ul><li><p>Local numbers starting with <code inline="">0</code> are converted to <code inline="">+&lt;country_code&gt;&lt;number&gt;</code> using the admin-configured default country code.</p></li><li><p>International numbers starting with <code inline="">+</code> are passed through unchanged.</p></li><li><p>Numbers using the <code inline="">00</code> international prefix are passed through unchanged.</p></li></ul><h2>Testing</h2><ul><li><p><strong>Unit tests</strong> in <code inline="">sms.rs</code> cover:</p><ul><li><p>E.164 validation</p></li><li><p>Environment-variable block detection</p></li></ul></li><li><p><strong>Integration test</strong> <code inline="">dashboard_bookings_shows_guest_phone_for_sms_booking</code> in <code inline="">web/mod.rs</code> verifies that the guest's phone number is displayed in the bookings dashboard when SMS is enabled.</p></li></ul></body></html>
</body>
</html>

## Summary by CodeRabbit

- **New Features**
  - Added optional SMS notifications for booking confirmations, cancellations, approvals, and host cancellations.
  - Added administrator controls for SMS setup, enablement, default country code, test messages, and configuration clearing.
  - Added event-level SMS opt-in settings.
  - Booking forms now accept optional phone numbers with validation and normalization.
  - Dashboards display guest phone numbers when available.
  - Added support for environment-managed SMS configuration.

### DIFF
--- a/migrations/062_sms_notifications.sql
+++ b/migrations/062_sms_notifications.sql
@@ -1,0 +1,22 @@
+-- SMS notifications via Twilio (optional, opt-in per event type).
+--
+-- Mirrors the smtp_config pattern: a system-wide singleton row, auth token
+-- encrypted at rest with AES-256-GCM (see src/crypto.rs), editable from the
+-- admin panel or CALRS_TWILIO_* environment variables.
+
+-- Guests can optionally leave a phone number on the booking form. It is only
+-- collected/shown when the event type has SMS notifications enabled.
+ALTER TABLE bookings ADD COLUMN phone_number TEXT;
+
+-- Per-event-type opt-in switch (default off, so existing event types keep
+-- working exactly as before with no SMS involved).
+ALTER TABLE event_types ADD COLUMN sms_notifications_enabled INTEGER NOT NULL DEFAULT 0;
+
+CREATE TABLE IF NOT EXISTS twilio_config (
+    id                TEXT PRIMARY KEY,
+    account_sid       TEXT NOT NULL,
+    auth_token_enc     TEXT,
+    from_number       TEXT NOT NULL,
+    enabled           INTEGER NOT NULL DEFAULT 1,
+    created_at        TEXT NOT NULL DEFAULT (datetime('now'))
+);
--- a/migrations/063_sms_default_country_code.sql
+++ b/migrations/063_sms_default_country_code.sql
@@ -1,0 +1,2 @@
+-- Default country calling code used to normalize local guest phone numbers.
+ALTER TABLE twilio_config ADD COLUMN default_country_code TEXT NOT NULL DEFAULT '+1';
--- a/src/db.rs
+++ b/src/db.rs
@@ -841,6 +841,7 @@ mod tests {
             "sessions",
             "auth_config",
             "smtp_config",
+            "twilio_config",
             "groups",
             "user_groups",
             "event_type_calendars",
@@ -875,7 +876,7 @@ mod tests {
             .fetch_one(&pool)
             .await
             .unwrap();
-        assert_eq!(count.0, 61, "All 61 migrations should be tracked");
+        assert_eq!(count.0, 63, "All 63 migrations should be tracked");
     }
 
     #[tokio::test]
@@ -889,7 +890,7 @@ mod tests {
             .fetch_one(&pool)
             .await
             .unwrap();
-        assert_eq!(count.0, 61, "Still 61 migrations after second run");
+        assert_eq!(count.0, 63, "Still 63 migrations after second run");
     }
 
     #[tokio::test]
--- a/src/db.rs
+++ b/src/db.rs
@@ -260,6 +260,14 @@ pub async fn migrate(pool: &SqlitePool) -> Result<()> {
             "061_booking_unique_per_member",
             include_str!("../migrations/061_booking_unique_per_member.sql"),
         ),
+        (
+            "062_sms_notifications",
+            include_str!("../migrations/062_sms_notifications.sql"),
+        ),
+        (
+            "063_sms_default_country_code",
+            include_str!("../migrations/063_sms_default_country_code.sql"),
+        ),
     ];
 
     let mut applied_count = 0u32;
--- a/src/main.rs
+++ b/src/main.rs
@@ -20,6 +20,7 @@ mod providers;
 mod resources;
 mod rrule;
 mod settings;
+mod sms;
 mod utils;
 mod web;
 
--- a/src/models.rs
+++ b/src/models.rs
@@ -144,6 +144,11 @@ pub struct EventType {
     /// Minimum notice (in minutes) required for a guest-initiated reschedule.
     /// `None` or `Some(0)` means no restriction.
     pub reschedule_notice_min: Option<i32>,
+    /// Opt-in: when true, the booking form shows an optional phone number
+    /// field and SMS notifications (via Twilio) are sent for this event
+    /// type's bookings, in addition to email. Defaults to false so existing
+    /// event types are unaffected.
+    pub sms_notifications_enabled: bool,
 }
 
 #[derive(Debug, Clone, FromRow, Serialize, Deserialize)]
@@ -196,6 +201,9 @@ pub struct Booking {
     pub reschedule_token: String,
     pub created_at: String,
     pub assigned_user_id: Option<String>,
+    /// Optional guest phone number in E.164 format (e.g. `+15551234567`),
+    /// collected only when the event type has SMS notifications enabled.
+    pub phone_number: Option<String>,
 }
 
 #[derive(Debug, Clone, FromRow, Serialize, Deserialize)]
--- a/src/sms.rs
+++ b/src/sms.rs
@@ -1,0 +1,494 @@
+//! SMS notifications via the Twilio REST API.
+//!
+//! Mirrors `email.rs`'s SMTP pattern on purpose: a system-wide singleton
+//! config, `CALRS_TWILIO_*` environment variables taking precedence over a
+//! DB-stored, AES-256-GCM-encrypted (`crate::crypto`) row editable from the
+//! admin panel. SMS is entirely opt-in — with no Twilio config *and* no
+//! event type enabling it, nothing changes for existing installs.
+//!
+//! Unlike SMTP, sending happens over plain HTTPS (Twilio's REST API), so we
+//! reuse the `reqwest` client already used for CalDAV instead of pulling in a
+//! Twilio SDK.
+
+use anyhow::{bail, Context, Result};
+use sqlx::SqlitePool;
+
+pub struct TwilioConfig {
+    pub account_sid: String,
+    pub auth_token: String,
+    pub from_number: String,
+}
+
+impl std::fmt::Debug for TwilioConfig {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("TwilioConfig")
+            .field("account_sid", &self.account_sid)
+            .field("auth_token", &"<redacted>")
+            .field("from_number", &self.from_number)
+            .finish()
+    }
+}
+
+pub struct TwilioStatus {
+    pub account_sid: String,
+    pub from_number: String,
+    pub enabled: bool,
+    pub from_env: bool,
+}
+
+const TWILIO_ENV_VARS: &[&str] = &[
+    "CALRS_TWILIO_ACCOUNT_SID",
+    "CALRS_TWILIO_AUTH_TOKEN",
+    "CALRS_TWILIO_FROM_NUMBER",
+];
+
+/// Read a Twilio env var, returning `Some(value)` only when set and non-empty.
+fn optional_twilio_env(name: &str) -> Option<String> {
+    std::env::var(name)
+        .ok()
+        .filter(|value| !value.trim().is_empty())
+}
+
+/// Load Twilio config from the `CALRS_TWILIO_*` environment block. Same
+/// "full block override" semantics as `email::load_smtp_config_from_env`:
+/// a complete block wins over the database; a partial block is ignored and
+/// the caller falls back to the DB config.
+fn load_twilio_config_from_env() -> Option<TwilioConfig> {
+    if !TWILIO_ENV_VARS
+        .iter()
+        .any(|name| std::env::var_os(name).is_some())
+    {
+        return None;
+    }
+
+    match (
+        optional_twilio_env("CALRS_TWILIO_ACCOUNT_SID"),
+        optional_twilio_env("CALRS_TWILIO_AUTH_TOKEN"),
+        optional_twilio_env("CALRS_TWILIO_FROM_NUMBER"),
+    ) {
+        (Some(account_sid), Some(auth_token), Some(from_number)) => Some(TwilioConfig {
+            account_sid,
+            auth_token,
+            from_number,
+        }),
+        _ => {
+            tracing::warn!(
+                "partial CALRS_TWILIO_* environment block (missing one of ACCOUNT_SID/AUTH_TOKEN/FROM_NUMBER); falling back to database Twilio config"
+            );
+            None
+        }
+    }
+}
+
+/// Whether the `CALRS_TWILIO_*` environment block governs the config (locks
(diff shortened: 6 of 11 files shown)


olivierlambert:
Thanks a lot for this, Cristian. It reads well and following the existing SMTP pattern was the right call, it made the whole thing easy to review.

Before merging I would like to generalise the provider side. Rather than binding the schema, the env vars and the admin panel to Twilio, I want a small `SmsProvider` trait with one adapter per gateway (the same shape as the existing `CalendarProvider` abstraction in `src/providers/`), so Twilio, GatewayAPI, seven.io and a generic webhook can all plug in without another migration later. Your Twilio client, the per-event-type opt-in, the phone normalisation and the booking form plumbing are all reused for it.

I am doing that work on a dedicated branch built on top of your commit, so authorship and credit stay with you. I will link it here as soon as it is pushed.

One thing I want to be upfront about: I am not going to merge this quickly. SMS costs money per message and it sits right in the booking path, so I want real users to field-test it on a branch before it lands in a release. If you are able to keep testing the Twilio path against your own account once the branch is up, that would help a lot, and any feedback from your side on the generalised config would be very welcome.

olivierlambert:
Branch is up: #181.

Your commit is the base of it, so the Twilio client, the per-event-type opt-in, the phone normalisation and the booking form plumbing are all yours in the history.

What changed on top: `SmsProvider` trait plus a factory and a provider registry, four adapters (Twilio, GatewayAPI, seven.io, and a generic webhook so any other gateway is a small script rather than a feature request), one generic `sms_config` table instead of `twilio_config`, `CALRS_SMS_*` instead of `CALRS_TWILIO_*`, and a single admin card that relabels itself per gateway. I also localised the guest-facing strings and the message bodies, added the reschedule and reminder messages that were missing, and dropped the phone-input JS since the server normaliser already covers those cases.

If you can test the Twilio path against your account on that branch, that would be the most useful thing right now, especially the approval flow and the new reminder message. Same for anyone else reading this with a GatewayAPI or seven.io account: those two adapters are written from the documentation and have unit tests, but they have never touched a live gateway.

Chr1s16:
Thank you, Olivier! This looks great, and I agree the provider abstraction makes a lot of sense, I was thinking initially that having just Twilio makes things simpler but you are totally right, in the future this can mean people depend on one single provider and that's not really feasible.

I’ll check out #181 and test the Twilio path against my account, especially the approval flow and the new reminder message. I’ll let you know if I run into anything or notice anything that needs adjusting. Happy to help!

olivierlambert:
Hey @Chr1s16, the generalised branch is up: #181.

Your commit is the base of it, so the Twilio client, the per-event-type opt-in, the phone normalisation and the booking form plumbing are all yours in the history. Close this PR whenever suits you, or leave it open until #181 lands. That one references this one either way, so nothing gets lost.

If you have the time and the Twilio credit to spare, testing is genuinely the thing standing between this and a release. Here's what would help most.

Grab the branch with `git fetch origin feat/sms-notifications && git checkout feat/sms-notifications && cargo build --release`. It targets `i18n` rather than `main` because it adds guest-facing strings and the translators get a round first. There's a full setup page at `docs/src/sms.md` on the branch if anything below is unclear.

Configuration lives in Dashboard > Admin > SMS settings now. Pick Twilio, drop in your Account SID, auth token and sender, and you're set. Two settings are new: a default country code, which is what turns a locally typed number into E.164, and a daily message limit you can leave at 0 for now. The Test gateway button sends a real message if you give it a number, or just checks your credentials for free if you leave the field empty. One gotcha if you were using environment variables: `CALRS_TWILIO_ACCOUNT_SID` and friends are now `CALRS_SMS_PROVIDER=twilio`, `CALRS_SMS_API_KEY`, `CALRS_SMS_API_SECRET` and `CALRS_SMS_SENDER`.

The event type setting is a three-state dropdown now (off, optional, required) rather than a checkbox. Set it to optional and the interesting run is:

1. Book with a number and check the confirmation arrives.
2. Same thing on an event type that requires confirmation. Nothing should arrive at booking time, and the SMS should turn up when you approve from the host email.
3. Reschedule it, both from the guest link and host-initiated, and check the "moved" message.
4. Cancel it, both as the guest and as the host, and check the "cancelled" message.
5. Set a short reminder time and let the background loop fire.

Steps 3 and 5 didn't exist in your PR, so nobody has actually watched those two arrive yet. They're the ones I'd most like a second pair of eyes on.

A few other things worth poking at. An optional-mode booking with the phone field left empty should just go through, no SMS and no complaint, while required-mode should block it with an inline error in the field rather than throwing away the whole form. A local number like 06 12 34 56 78 with the country code set to +33 should land in the database as +33612345678 and show up next to the guest email on the bookings dashboard. Saving the settings with the token field empty should keep your stored token. Switching the gateway dropdown to seven.io and saving with an empty secret should be refused rather than quietly reusing your Twilio token as a seven.io key. And if you set the daily limit to 1 and make two bookings, the second should still send its email but no SMS.

The most useful thing to hear back is any error Twilio returns that we categorise wrongly. Failures get normalised into auth, invalid recipient, invalid sender, insufficient credit and rate limited, so if the panel reports the wrong one, paste the raw error and I'll fix the mapping. After that, whether the segment count and cost show up in the admin card once a message goes out, and whether the wording reads naturally to you.

No obligation at all, and thanks again for the original PR. It saved a lot of groundwork.

Chr1s16:
Thanks again, Olivier, for putting so much thought and work into this, and for being so open to feedback and testing. I really appreciate the effort you’re putting into making this solid before it lands in a release.

I’ve been testing the Twilio path locally with a Twilio Trial account, since I don’t currently have a paid Twilio account, and I came across one thing that might be useful to support explicitly: a small Twilio trial mode.

Twilio Trial accounts have a few restrictions, mainly around verified recipients and the message body. I added a small `CALRS_SMS_TWILIO_TRIAL=true` switch locally which uses Twilio’s predefined `sms_appointment_reminders` template instead of the custom message body. This makes it possible to test the actual Twilio API integration and verify that SMS messages are sent and received correctly across the different booking flows.

I’ve tested it locally with regular bookings, reminders, cancellations, and team bookings, and all the SMS messages are being sent and received correctly using the Trial template.

The nice thing is that it’s completely opt-in: if someone wants to use a regular paid Twilio account, they can simply set `CALRS_SMS_TWILIO_TRIAL=false` or remove the variable entirely, and the normal message bodies are used again.

Do you think this would be useful to include? If so, I’d be happy to adapt it to the new `SmsProvider` abstraction on #181 rather than opening a separate PR. I think it could make the Twilio path much easier for contributors to test without needing a paid account.

olivierlambert:
This is a genuinely good catch, thank you. The Twilio path being untestable without a paid account was a real gap, and I had not appreciated that trial accounts refuse custom bodies outright rather than just restricting recipients. That is a much bigger obstacle for contributors than the verified-number limit.

I have gone ahead and implemented it on #181 (`8116ddf`) so it lands with the rest of the provider work rather than trailing it. The idea and the field testing are yours, and the commit says so.

It works the way you described. `CALRS_SMS_TWILIO_TRIAL=true` substitutes `sms_appointment_reminders` for the composed message, so the credentials, the sender, the recipient normalisation, the response parsing and all four booking events run against the real API. A few things I tightened on top of your version:

**Environment only.** No `sms_config` column and no admin field. An operator who could flip it from the panel would eventually flip it and ship canned templates to real guests. It also sits outside the all-or-nothing `CALRS_SMS_*` block and is read on its own, so it works alongside a database-stored config rather than forcing you to move everything to the environment just to test.

**Confined to `twilio.rs`.** `SmsProvider`, `factory.rs` and `message.rs` are untouched, so the other three gateways do not grow a Twilio-shaped concept they have no use for.

**Loud while active.** A warning on every send plus a badge on the admin SMS card. Once the template is substituted, `Confirmed`, `Cancelled`, `Rescheduled` and `Reminder` are indistinguishable on the handset, so the log is the only place to tell which one fired. The composed body is still built and logged at debug level, which keeps `message.rs` and the segment estimation on the exercised path in the mode most contributors will actually run.

**A guard on the expensive direction.** Red-teaming my own change turned up one way this can cost money rather than save it: on a *paid* account the template name is just text, so a flag left set after an upgrade would text every guest the literal string `sms_appointment_reminders` at full price. "Test gateway" with the recipient left empty already fetches the account, so it now reads the account type and refuses outright when the flag is set on a `Full` account. The opposite mistake needs no guard, it fails closed at send time with Twilio's own refusal.

`request_body()` and `trial_mismatch()` are pure functions with unit tests, same shape as `parse_error`, no mock server. Documented in `docs/src/sms.md` under a heading that says plainly it is a testing aid rather than a deployment option.

To answer the question I was going to ask you: it turns out `Body` carries the template *name* directly rather than a `ContentSid`, so the request shape is identical on both paths and `SendReceipt` parsing needed no changes at all. That made this much smaller than I expected.

Could you re-test with your trial account now that it is on the branch? That would let us validate the whole PR end to end without a paid account, which is exactly what you were getting at. The paths I would most like covered:

- All four events: booking, reminder, cancellation by the guest and by the host, plus a reschedule.
- An event type set to `required` phone mode, and one set to `optional` left empty, confirming the second sends no SMS and does not break the booking.
- A booking on a team event type, since the round-robin assignment picks the host after the phone is collected.
- "Test gateway" with the recipient left empty, which should report the account as trial and accept.
- The daily cap: set `CALRS_SMS_DAILY_CAP=1` and confirm the second booking still completes and still emails, just without the SMS.

If any of that misbehaves I would rather find it now than after a release. Thanks again for pushing on this.

Chr1s16:
Just finished testing everything with my Twilio Trial account, and it works great! I tested every point you mentioned multiple times, including all the booking flows, and everything behaves exactly as it should.

gsmachado:
It would be nice to merge this. This is a good feature! Thanks for spending your attention and your tokens on this. 😅

olivierlambert:
I just wanted a bit more feedback on it. Have you tested it @gsmachado ?

gsmachado:
@olivierlambert everything seems to work, yes. I have a productive Twilio account, and I tested it: I get an SMS once I book something.

However, before merging, something needs to be fixed on this screen (which is related to user experience):

image

The `Phone number` needs to be flagged as optional with `(optional)` -- i.e., follow the same pattern as with the `Notes`.

Other than that, I think it's ready to go.

Well, of course, there are some things to improve, but I'd rather leave it to another PR/issue. For example, add an option (in the admin panel) to minimally customize the SMS text.

Chr1s16:
@gsmachado You can set it to optional in the Event type settings under Booking options. I think this is the best option since if you ever require to deploy it somewhere where people must put in their phone number it is more effective this way, having (optional) as only option might make them skip that step. 

image

gsmachado:
@Chr1s16 have you pushed the latest changes to your branch? I can't see any dropdown related to SMS notifications. Just the checkbox.

Chr1s16:
@gsmachado oh, you tested this one. There's an updated version with the dropdown menu + multiple sms gateways and a genereic webhook: https://github.com/olivierlambert/calrs/pull/181

gsmachado:
@Chr1s16 I see!

Ok, now I tested #181, and it works fine with a real Twilio account.

@olivierlambert now, yes, it's ready to merge. 👍

olivierlambert:
Your Twilio work shipped to `main` today in #184, with your commit `e8354bf` as the base of the branch, so I'm closing this one as superseded rather than stale.

What it grew into, on top of what you wrote: the Twilio client moved behind an `SmsProvider` trait (same shape as the existing `CalendarProvider`), and three more gateways joined it, GatewayAPI, seven.io, and a generic HMAC-signed webhook so a gateway with no adapter is a small script away. Each adapter owns its own response parsing, because "the credentials are wrong" turns out to look different everywhere: HTTP 401 on Twilio, 401 with a `{"code": "0x0213"}` body on GatewayAPI, and HTTP **200** carrying `{"success": "900"}` on seven.io.

The trial-account mode is yours specifically, and it came directly out of your field testing: `CALRS_SMS_TWILIO_TRIAL=true` substitutes Twilio's `sms_appointment_reminders` template so the whole path is exercisable without a paid account. Thank you for finding that, it's the difference between the Twilio adapter being testable by contributors and not.

Two things worth knowing if you run it:

- With the flag on, all four events look identical on the handset, since the template replaces the composed body. There's a warning on every send and a badge in the admin SMS card for exactly that reason.
- Setting it on a *full* account is the expensive mistake, so `check()` reads the account type and refuses. Note that guard only fires on the free credential check behind "Test gateway" with the recipient left empty, so click that first after changing the flag.

If you get a chance to run `main` against your Twilio account and confirm the four events (confirmed, cancelled, rescheduled, reminder) land as expected, that field test would be genuinely useful. Thanks again.

PR #186, opened 2026-08-19
feat(event-types): per-event-type rolling booking horizon

iamthew4lrus789:
Event types currently have no upper bound on how far ahead a guest can book — someone can open a booking page and schedule eleven months out. This adds an optional per-event-type rolling horizon: a "Sales intro" can stay bookable 14 days ahead while an "Annual review" stays open indefinitely.

There is no upstream issue for this; I raised it myself, so please say if the shape isn't what you'd want.

## Schema

`migrations/063_booking_horizon.sql`, mirroring `050_min_notice_cancel_reschedule.sql`:

```sql
ALTER TABLE event_types ADD COLUMN booking_horizon_days INTEGER;
```

**`NULL` is unlimited**, so every existing row keeps today's behaviour with no data migration. `0` means today only. Registered in the `migrations` array in `src/db.rs::migrate()` immediately after `062_sms_notifications`, and verified applied against a real DB (`SELECT name FROM _migrations WHERE name LIKE '063%'` plus `PRAGMA table_info(event_types)`).

One column, no mode enum, no fixed end-date variant — deliberately the same shape as migration 050.

## Semantics

A slot whose local start date is `D` is offerable iff `D <= today_in_host_tz + booking_horizon_days`, inclusive at the boundary. "Today" is derived in the event type's own timezone using the existing `now_host` pattern rather than UTC, so hosts either side of the date line don't get off-by-one reports.

## Enforcement — four kinds of site

A slot list is advisory, so hiding slots is not the limit:

1. **Slot computation.** `compute_slots` threads the horizon into `compute_slots_from_rules`, which clamps its day loop. `day_offset` already counts days from today in the host timezone, so it compares directly against the horizon. This covers all five slot-rendering handlers at once (`show_group_slots`, `show_dynamic_group_slots`, `show_slots_for_user`, `show_slots`, `guest_reschedule_slots`).
2. **Booking POSTs.** All five booking handlers (`handle_group_booking`, `handle_dynamic_group_booking`, `handle_booking_for_user`, `handle_booking`, `guest_reschedule_booking`) reject a beyond-horizon datetime beside their existing `min_notice` guard. This is the actual enforcement — a crafted POST cannot cross it.
3. **Troubleshoot.** `troubleshoot()` reimplements availability for a single date and shares no code with `compute_slots`, so it gets its own check, surfaced as a `beyond_horizon` blocking reason the way `resource_busy` already is.
4. **Month navigation.** `build_month_params` now returns `next_month: Option<String>`, `None` once the following month starts past the horizon, so the forward arrow disappears instead of leading to an empty month.

Minimum notice and the horizon are independent. When they cross, the event type has no bookable slots and the page renders its existing empty state — covered by a test.

## Both form paths are covered

`booking_horizon_days` is wired through `EventTypeForm` and **both** the personal and team paths — create defaults, INSERT, edit prefill, and UPDATE on each side. I checked this specifically because #114 was "persist booking frequency limits on team event types", i.e. exactly this gap shipping once already. There's a test asserting the horizon survives a **team** form round-trip.

One thing worth flagging: `event_type_form.html` actually has **five** render sites, not four. The fifth is `render_event_type_form_error()`, the re-render after a validation error. Without it the user's entered horizon silently blanks back to "no limit" on every failed save, so it's wired too.

## CLI

`EventTypeCommands` has no `set`/`update` subcommand and `Create` exposes no notice or limit flags, so I added no CLI setters — that would be a new pattern rather than parity. `event-type slots` does respect the horizon: `--days` is clamped to it so the CLI shows the same window as the booking page, and the help text says so.

## i18n

New keys are in `i18n/en/main.ftl` only (`event-type-form-booking-horizon-label`, `-help`), following the surrounding `event-type-form-*` convention and the register of "Leave empty for no restriction." Other locales fall back to English at runtime.

Because this adds UI text, the PR targets **`i18n`, not `main`**, per CLAUDE.md: *"New features that touch UI text should branch off `i18n`, not `main`"* — going straight to `main` would leave translators out of sync.

I left the hard-coded English "Minimum notice" label at its existing site alone; it's a pre-existing inconsistency and not this PR's to fix.

## CHANGELOG

Left untouched — every heading there is a released version, so the entry belongs to your release process, not to this branch.

## Verification

All four CI jobs run locally against a pinned `rust:latest` container (rustc 1.97.1):

| Job | Result |
|---|---|
| `cargo fmt --check` | pass |
| template `{% block %}` balance | pass (40 templates) |
| `RUSTFLAGS="-D warnings" cargo clippy --profile=test` | pass |
| `RUSTFLAGS="-D warnings" cargo test` | pass — 878 passed, 0 failed (18 new) |

No new `#[allow(...)]`; the crate-wide allows in `main.rs` already cover the argument count.

Tests cover: `NULL` behaves exactly as today; the last in-horizon day is offered and the first out-of-horizon day is not; `0` yields today only; each of the five booking POSTs rejects a beyond-horizon datetime; `build_month_params` returns `next_month: None` at the boundary; a horizon on a team event type survives a form round-trip; and `min_notice_min` exceeding the horizon yields zero slots without error. The POST-guard tests each carry a control that re-runs the identical request with the horizon cleared and asserts the booking *is* created, so they can't pass vacuously.

Beyond the suite I ran a real instance with three event types — horizon 14, horizon 0, and no horizon — and checked the booking pages on 2026-08-19 (UTC):

| Event type | Sept view: dates offered | Forward arrow |
|---|---|---|
| horizon 14 | `2026-09-01 .. 2026-09-02` (stops exactly at today + 14) | disabled |
| horizon 0 | none | disabled |
| no horizon | `2026-09-01 .. 2026-10-01` | live |

`event-type slots --days 30` matches: 14 → last day `2026-09-02`, none → `2026-09-17`.

## One pre-existing bug found, deliberately not fixed

While making `next_month` optional I hit a minijinja detail: an `Option::None` passed into a template renders as the literal string `"none"`, and `| default(value='')` does not catch it because the value is *defined*, just none. My `nextMonth` emits `""` correctly.

**`prevMonth` has the same bug on `main` today** — `prev_month` was already `Option<String>`, so `slots.html` emits `"prevMonth": "none"` when viewing the current month. In `updateNavArrows()` the guard is `if (data.prevMonth)`, and `"none"` is truthy in JS, so after an AJAX month navigation the back arrow renders even when there is no previous month, linking to `?month=none`. The server-rendered path is fine because `{% if prev_month %}` treats `None` as falsy.

I've left it alone — it's unrelated to this feature and fixing it changes back-arrow behaviour, which deserves its own PR rather than being buried here. Happy to send that separately if useful.

## Summary by CodeRabbit

- **New Features**
  - Added configurable booking horizons for event types, including unlimited, today-only, or a set number of days ahead.
  - Enforced booking limits across scheduling, rescheduling, calendars, and CLI slot searches.
  - Calendar navigation now stops when no dates remain within the booking horizon.
  - Added clear “beyond booking horizon” status information in troubleshooting views.
  - Added booking horizon controls to event type settings.

- **Documentation**
  - Documented booking horizon configuration, timezone behavior, enforcement, and CLI effects.

### DIFF
--- a/docs/src/cli-reference.md
+++ b/docs/src/cli-reference.md
@@ -61,10 +61,12 @@ calrs event-type create [OPTIONS]
 calrs event-type list
 
 calrs event-type slots <SLUG> [OPTIONS]
-    --days <DAYS>    Number of days to show (default: 7)
+    --days <DAYS>    Number of days to show (default: 7, clamped to the booking horizon)
 ```
 
 `slots` consults any [shared resources](./resources.md) attached to the event type: slots where a required resource is busy are not printed, matching the web booking page.
+
+If the event type sets a [booking horizon](./event-types.md#booking-horizon), `--days` is clamped to it, so the CLI shows the same window the booking page offers. Asking for `--days 30` on an event type with a 5-day horizon prints 6 days (today plus five).
 
 ### `calrs booking`
 
--- a/docs/src/event-types.md
+++ b/docs/src/event-types.md
@@ -107,6 +107,22 @@ Enable **"Limit booking frequency"** to cap how many bookings can be made per ti
 
 Both settings are configured via toggle switches in the **Booking limits** card of the event type form.
 
+## Booking horizon
+
+**Booking horizon** caps how far into the future a guest may book, as a rolling window of days. A "Sales intro" might stay bookable 14 days ahead while an "Annual review" stays open indefinitely.
+
+The field lives in the **Buffers & notice** card of the event type form:
+
+| Value | Meaning |
+|---|---|
+| *(empty)* | No limit — guests can book arbitrarily far ahead. This is the default and the existing behaviour. |
+| `0` | Today only. |
+| `14` | Today plus the next 14 days, inclusive. |
+
+The window is measured in the event type's own timezone, so hosts either side of the date line get the day they expect. The month arrow on the booking page disappears once the next month starts past the horizon, and `calrs event-type slots` clamps its `--days` window to match.
+
+The limit is enforced when the booking is submitted, not just when slots are drawn, so a crafted request cannot book past it. Minimum notice and the horizon are independent: if minimum notice pushes the earliest bookable slot past the horizon, the event type simply has no bookable slots and the page shows its normal empty state.
+
 ## Calendar views
 
 The guest slot picker supports three views, switchable via icons in the calendar header:
@@ -142,7 +158,8 @@ Available slots are computed by:
 4. Filtering out slots blocked by required [shared resources](./resources.md) (all mode: any busy resource blocks; round-robin mode: blocked only when every resource is busy)
 5. Applying buffer times (before and after each slot)
 6. Removing slots that violate minimum notice (too close to now)
-7. If "one slot per day" is enabled, keeping only the earliest slot per day
+7. Removing days past the booking horizon (too far ahead)
+8. If "one slot per day" is enabled, keeping only the earliest slot per day
 
 ```bash
 # View available slots for the next 7 days
--- a/i18n/en/main.ftl
+++ b/i18n/en/main.ftl
@@ -275,6 +275,8 @@ event-type-form-resources-mode-round-robin = Any one free resource is enough (it
 event-type-form-notice-unit-minutes = minutes
 event-type-form-notice-unit-hours = hours
 event-type-form-notice-unit-days = days
+event-type-form-booking-horizon-label = Booking horizon
+event-type-form-booking-horizon-help = How many days ahead guests can book. Leave empty for no limit, 0 for today only.
 
 # Booking confirmation: cancel/reschedule policy notices (templates/confirmed.html)
 
--- a/migrations/063_booking_horizon.sql
+++ b/migrations/063_booking_horizon.sql
@@ -1,0 +1,3 @@
+-- Per-event-type rolling booking horizon: how many days into the future a
+-- guest may book. NULL means unlimited (the existing behaviour).
+ALTER TABLE event_types ADD COLUMN booking_horizon_days INTEGER;
--- a/src/commands/event_type.rs
+++ b/src/commands/event_type.rs
@@ -38,10 +38,20 @@ pub enum EventTypeCommands {
     Slots {
         /// Event type slug
         slug: String,
-        /// Number of days to show
+        /// Number of days to show (clamped to the event type's booking horizon)
         #[arg(long, default_value = "7")]
         days: i32,
     },
+}
+
+/// Clamp a requested day count to an event type's rolling booking horizon.
+/// A horizon of N days leaves days 0..=N bookable, so it caps the window at
+/// N + 1 days. `None` means unlimited and leaves the request untouched.
+fn clamp_days_to_horizon(days: i32, horizon: Option<i32>) -> i32 {
+    match horizon {
+        Some(h) => days.min(h.saturating_add(1)),
+        None => days,
+    }
 }
 
 #[derive(Tabled)]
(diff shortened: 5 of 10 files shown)


olivierlambert:
Merged, and already on `main`: `i18n` and `main` were level, so this fast-forwarded straight through as `a87fafe`.

Thanks for this. It is one of the most carefully scoped contributions this repo has had. Enforcing at the POST handlers and not just the slot list is the right call, and catching the fifth render site in `render_event_type_form_error()` is exactly the gap that already shipped once as #114. The control cases on the POST-guard tests, re-running the identical request with the horizon cleared, are what made the suite worth trusting rather than just green.

Verified locally before merging, on a clean worktree of your head:

| Gate | Result |
|---|---|
| `cargo fmt --check` | pass |
| `RUSTFLAGS="-D warnings" cargo clippy --profile=test` | pass |
| `cargo test` | 880 passed, 0 failed |

On the three CodeRabbit findings, for the record:

- The overflow panic in `horizon_last_date` is a false positive. I probed `0`, `1`, `365`, `95_005_570`, `i32::MAX - 1` and `i32::MAX` through both `horizon_last_date` and `build_month_params`, plus negatives. Nothing panics. `checked_add_signed` returns `None` past chrono's year 262143 and every call site reads that as unlimited, which is what your doc comment already says.
- The `CHECK` constraint is reasonable and would match the seven migrations that use one, but `parse_optional_day_count` already filters negatives and only direct SQL could write one. Not worth holding the merge.
- The CLI timezone point is real but pre-existing: `event-type slots` already derives everything from `Local::now()`, including `min_notice`. You inherited it rather than introduced it.

Your `prevMonth` finding is real and I confirmed it: minijinja renders `None` as the literal string `"none"` and `| default(value='')` does not catch it, because the value is defined. So `slots.html` emits `"prevMonth": "none"` on the current month and the guard at line 1211 is truthy for it, while line 1557 in the same file already uses the correct `hasPrevMonth && prevMonth` shape. Cosmetic, since `parse_month_param` falls back to the current month, but genuinely inconsistent. Please do send that separate PR, and you were right to keep it out of this one.

If you run this in anger, I would value a note on how the horizon behaves for you across a DST boundary and with `min_notice` set close to it. That is the corner where field use tends to find things a suite does not.

olivierlambert:
Small change of plan on the `prevMonth` bug: I had a window free this afternoon, so I went ahead and did it myself rather than leave it sitting on your plate. Sorry for taking the occasion after inviting you to send it, no reflection on you at all. It's #187.

Your diagnosis was exactly right, including the part about `default(value='')` not catching it because the value is defined rather than undefined. I confirmed it by rendering the template with `prev_month` as `None` and it does emit `"prevMonth": "none"` verbatim.

Two things I added on top of what you described:

- The payload fix alone is sufficient, since an empty string is falsy and the existing `if (data.prevMonth)` guard then behaves. I gated on `hasPrevMonth` as well anyway, so the back arrow stays symmetric with the forward arrow you did in #186.
- I swept the other seven `default(value=)` sites in the templates to see whether the same class was hiding elsewhere. `guest_date` is a `String` and the rest resolve to `&str` via `unwrap_or`, so `prev_month` was the only `Option` reaching one. Good news, it was isolated.

I also folded in the negative-horizon case from the review. `horizon_last_date` returned `None` on a large negative because the arithmetic ran off the representable calendar, and `None` reads as unlimited everywhere, so it failed open. Only reachable by writing the column with SQL directly, but that's the wrong direction to fail in. Your day loop in `compute_slots_from_rules` already handled it correctly, since `start_offset` is clamped to `>= 0`, so it was only the POST guard and troubleshoot that disagreed.

I left the CLI timezone point alone. You were right that it isn't this feature's to fix, and it's bigger than it looks: `event-type slots` derives everything from `Local::now()` and the machine timezone, not just the horizon. That one is genuinely up for grabs if you want it.

Thanks again for #186, and for flagging the `prevMonth` thing instead of quietly burying it in the same PR. That's the habit that makes review easy.

PR #191, opened 2026-08-21
fix(email): allow SMTP relays that advertise no AUTH

### LINKED ISSUE #190: bug: SMTP always authenticates, so an unauthenticated local relay can never send
## Summary

Hey, first thanks a lot for your project !

I think Calrs can't deliver through an SMTP server that advertises no AUTH mechanism. It's because `send_email()` attaches credentials to the lettre transport unconditionally, even when the configured username and password are empty, so a local MTA relaying anonymously from the loopback (Postfix, OpenSMTPD, Stalwart, Mailpit) fails every send with:

```
ERROR calrs::email: email delivery failed to=guest@example.com
error=internal client error: No compatible authentication mechanism was found
```

The configuration is representable: `calrs config smtp` accepts an empty username and an empty password, and so does the admin SMTP form. It simply can never send. Nothing in the CLI, the admin panel or the docs hints that credentials are mandatory.

Seen on v1.15.1, still present on `main` (ccae106).

## The root cause

[`src/email.rs:2129`](https://github.com/olivierlambert/calrs/blob/ccae1061ddb84d15a99885b5c1c2ff139b4505d3/src/email.rs#L2129), where the credentials are built and attached before the transport is, on both TLS branches:

```rust
async fn send_email(config: &SmtpConfig, email: Message) -> Result<()> {
    let creds = Credentials::new(config.username.clone(), config.password.clone());

    let mailer = match config.tls_mode {
        SmtpTlsMode::Tls => AsyncSmtpTransport::<Tokio1Executor>::relay(&config.host)?
            .port(config.port)
            .credentials(creds)
            .build(),
        SmtpTlsMode::StartTls => {
            AsyncSmtpTransport::<Tokio1Executor>::starttls_relay(&config.host)?
                .port(config.port)
                .credentials(creds)
                .build()
        }
    };
```

lettre only skips authentication when the transport got no credentials at all. Given `Credentials::new("", "")` it looks for a mechanism to use, finds that the server advertises none, and fails the send before `DATA`, so the SMTP server never sees the message. This is a client-side abort, not a rejection: the very same relay accepts a hand-written SMTP session from the same host.

The env-var path cannot express "no auth" either. `load_smtp_config_from_env()` (`src/email.rs:1883`) requires `CALRS_SMTP_USERNAME` and `CALRS_SMTP_PASSWORD` as part of the mandatory block, and leaving them out makes the block partial, which logs a warning and falls back to the database config.

## Perhaps the fix

Only attach credentials only when a username is set:

```rust
async fn send_email(config: &SmtpConfig, email: Message) -> Result<()> {
    let builder = match config.tls_mode {
        SmtpTlsMode::Tls => AsyncSmtpTransport::<Tokio1Executor>::relay(&config.host)?,
        SmtpTlsMode::StartTls => {
            AsyncSmtpTransport::<Tokio1Executor>::starttls_relay(&config.host)?
        }
    }
    .port(config.port);

    // An empty username means "no SMTP AUTH". A local MTA relaying from the
    // loopback (Postfix, OpenSMTPD, Stalwart, Mailpit) advertises no AUTH
    // mechanism at all, and lettre fails the send with "No compatible
    // authentication mechanism was found" rather than skipping authentication.
    let mailer = if config.username.is_empty() {
        builder.build()
    } else {
        builder
            .credentials(Credentials::new(
                config.username.clone(),
                config.password.clone(),
            ))
            .build()
    };
```

Both match arms return an `AsyncSmtpTransportBuilder`, so `.port()` can be hoisted out of the match and the conditional `.credentials()` applied after it. Authenticated setups are unaffected because a non-empty username keeps exactly today's behaviour.

Worth considering alongside it: relax the env block to require only `CALRS_SMTP_HOST` and `CALRS_SMTP_FROM_EMAIL`, treating `CALRS_SMTP_USERNAME` and `CALRS_SMTP_PASSWORD` as optional, so a container deployment pointed at a sidecar MTA can be configured entirely from the environment.

## Step to reproduce

Against any MTA that accepts anonymous relay from the loopback and advertises no AUTH on port 25 (I found this issue here with Stalwart, but Postfix's default setup and Mailpit behave the same):

1. `calrs config smtp --host mail.example.com --port 25 --username '' --from-email noreply@example.com`, with an empty password at the prompt.
2. `calrs config smtp-test you@example.com`
3. Result: `Error: internal client error: No compatible authentication mechanism was found`. Nothing reaches the MTA, no queue entry, no log line on the server side.
4. Same server, same host, by hand, and the envelope is accepted:

```console
$ printf 'EHLO test\r\nMAIL FROM:<noreply@example.com>\r\nRCPT TO:<you@example.com>\r\nQUIT\r\n' | nc 127.0.0.1 25
220 mail.example.com Stalwart ESMTP at your service
250-STARTTLS
...                          (no AUTH advertised)
250 2.1.0 OK
250 2.1.5 OK
```

## Impact

Hits every self-hoster who relays through a local MTA rather than a third-party provider. Bookings still succeed, so nothing looks broken from the outside: confirmations, reminders and host notifications simply never leave, and the only trace is a line in the server log.

Hope its help!

olivierlambert:
Closes #190. Thanks @ManUtopiK for the report, which was accurate down to the line number and shipped its own fix.

## The bug

`send_email()` attached credentials to the lettre transport unconditionally, so a configured-but-empty username still put the transport into the authenticating path. lettre gates on presence, not emptiness:

```rust
// lettre-0.11.19 src/transport/smtp/async_transport.rs:495
if let Some(credentials) = &self.info.credentials {
    conn.auth(&self.info.authentication, credentials).await?;
}
```

Given `Credentials::new("", "")` it goes looking for a mechanism, a local MTA relaying anonymously from the loopback advertises none, and `get_auth_mechanism` returns `None` — the reported error, raised client-side before the envelope ever reaches the server. The configuration was representable through both `calrs config smtp` and the admin form. It simply could never send.

## What changed

Credentials are attached only when the username is non-empty, keyed on a new `SmtpConfig::uses_auth()`. Authenticated setups keep exactly today's behaviour.

Two adjacent gaps are closed alongside it, because neither left a working path for the deployment this is about:

**The env block could not express "no auth".** `CALRS_SMTP_USERNAME` and `CALRS_SMTP_PASSWORD` were part of the mandatory set, so omitting them made the block partial and silently fell back to the database. Only `CALRS_SMTP_HOST` and `CALRS_SMTP_FROM_EMAIL` are required now. A password with no username is a hard error, matching how an invalid `PORT`/`TLS_MODE` is treated in an otherwise-complete block.

**TLS was still mandatory.** `starttls_relay()` sets `Tls::Required`, and calrs builds lettre with `tokio1-rustls-tls`, whose roots come from `webpki-roots` — the compiled-in Mozilla bundle, not the system trust store. So a relay offering no STARTTLS (Mailpit's default) or presenting a self-signed certificate (Debian's Postfix default snakeoil) stayed unreachable in every mode, and the auth fix alone would have moved those users from an auth error to a TLS error. A third `tls_mode`, `none`, sends unencrypted. It warns when combined with a username, or aimed at a host that is not loopback.

Surfaced in all three places the reporter noted were silent: the CLI prompt (which now also skips the password prompt when the username is empty), the admin form, and the docs.

No migration: `smtp_config.tls_mode` is unconstrained `TEXT`.

## Verification

End to end against a local MTA advertising neither AUTH nor STARTTLS, driving the real `calrs config smtp-test` through the env block:

| | Result | Server saw |
|---|---|---|
| No username | `✓ Test email sent!` | `EHLO \| MAIL FROM \| RCPT TO \| DATA \| QUIT` |
| Username set | `No compatible authentication mechanism was found` | `EHLO`, then the client hung up |

The second row reproduces the issue exactly, including the reporter's observation that nothing reaches the MTA. Plus 7 new unit tests over the auth decision, the relaxed env block, the new TLS mode round-trip, and loopback-host detection. 891 tests green, clippy clean.

## Note for the reporter

@ManUtopiK your Stalwart setup should be fixed by the credentials change alone, since it advertises STARTTLS. If the certificate is publicly trusted, leave `tls_mode` at `starttls` and just clear the username. `none` is there for the Postfix-snakeoil and Mailpit cases you mentioned. Field reports welcome either way.

## Summary by CodeRabbit

* **New Features**
  * SMTP configuration now supports unauthenticated relays without requiring a username or password.
  * Added an unencrypted SMTP transport option (`none`/`plaintext`) for compatible relay setups.
  * SMTP settings can now be configured with only a server host and sender address.
* **Bug Fixes**
  * Improved validation for incomplete or invalid SMTP authentication settings.
  * Added warnings when plaintext SMTP could expose credentials, especially over non-local connections.
* **Documentation**
  * Expanded CLI, web settings, and environment-variable guidance for relay configuration, TLS modes, and certificate trust limitations.

### DIFF
--- a/docs/src/cli-reference.md
+++ b/docs/src/cli-reference.md
@@ -157,6 +157,13 @@ SMTP credentials, the environment still wins and calrs relays unauthenticated.
 It logs a warning once at startup when that happens, because the admin form
 locks itself whenever the environment governs.
 
+The two credential variables are not interchangeable when only one is present.
+`CALRS_SMTP_PASSWORD` without `CALRS_SMTP_USERNAME` is an error: the password
+can never be sent, so the configuration contradicts itself. A username without a
+password is accepted and only warns, since a permissive relay may still
+authenticate on the username alone, though in practice it usually means the
+password never reached the process.
+
 The equivalent from the CLI, with no prompts:
 
 ```
--- a/docs/src/cli-reference.md
+++ b/docs/src/cli-reference.md
@@ -100,6 +100,7 @@ calrs config smtp [OPTIONS]
     --username <USERNAME>   SMTP username (omit for an unauthenticated relay)
     --from-email <EMAIL>    Sender email address
     --from-name <NAME>      Sender display name
+    --tls-mode <MODE>       starttls (default), tls, or none
 
 calrs config show           Display current configuration
 
@@ -149,6 +150,18 @@ CALRS_SMTP_HOST=localhost
 CALRS_SMTP_PORT=25
 CALRS_SMTP_TLS_MODE=none
 CALRS_SMTP_FROM_EMAIL=noreply@example.com
+```
+
+If the environment block sets no `CALRS_SMTP_USERNAME` while the database holds
+SMTP credentials, the environment still wins and calrs relays unauthenticated.
+It logs a warning once at startup when that happens, because the admin form
+locks itself whenever the environment governs.
+
+The equivalent from the CLI, with no prompts:
+
+```
+calrs config smtp --host localhost --port 25 --tls-mode none \
+  --username '' --from-email noreply@example.com --from-name calrs
 ```
 
 ### `calrs resource`
--- a/docs/src/cli-reference.md
+++ b/docs/src/cli-reference.md
@@ -97,7 +97,7 @@ Configure SMTP, authentication, and OIDC.
 calrs config smtp [OPTIONS]
     --host <HOST>           SMTP server hostname
     --port <PORT>           SMTP port (default: 587)
-    --username <USERNAME>   SMTP username
+    --username <USERNAME>   SMTP username (omit for an unauthenticated relay)
     --from-email <EMAIL>    Sender email address
     --from-name <NAME>      Sender display name
 
@@ -115,6 +115,40 @@ calrs config oidc [OPTIONS]
     --client-secret <SECRET>  Client secret
     --enabled <BOOL>          Enable/disable OIDC
     --auto-register <BOOL>    Auto-create users on first login
+```
+
+#### Relaying through a local MTA
+
+Leaving the username empty configures an unauthenticated relay, which is how you
+send through a local MTA (Postfix, OpenSMTPD, Stalwart, Mailpit). calrs then
+attaches no credentials at all, rather than authenticating with an empty
+username, which such a relay rejects because it advertises no AUTH mechanism.
+
+The TLS mode prompt accepts three values:
+
+| Mode | Transport |
+|---|---|
+| `starttls` (default) | Plain connection upgraded with STARTTLS, typically port 587 |
+| `tls` | Implicit TLS from the first byte, typically port 465 |
+| `none` | No encryption, typically port 25 |
+
+Use `none` only for a relay on the same machine. calrs validates certificates
+against a compiled-in Mozilla root bundle rather than the system trust store, so
+a relay presenting a self-signed or private-CA certificate (Debian's Postfix
+default, for one) cannot be reached with `starttls` or `tls` either, and `none`
+over the loopback is the way through. Mail sent this way, and any credentials
+sent with it, cross the connection in the clear; calrs logs a warning if you
+combine `none` with a username, or point it at a host that is not loopback.
+
+The same applies to the `CALRS_SMTP_*` environment block, where only
+`CALRS_SMTP_HOST` and `CALRS_SMTP_FROM_EMAIL` are required. A full local-relay
+configuration is:
+
+```
+CALRS_SMTP_HOST=localhost
+CALRS_SMTP_PORT=25
+CALRS_SMTP_TLS_MODE=none
+CALRS_SMTP_FROM_EMAIL=noreply@example.com
 ```
 
 ### `calrs resource`
--- a/src/commands/config.rs
+++ b/src/commands/config.rs
@@ -16,7 +16,7 @@ pub enum ConfigCommands {
         /// SMTP port (default: 587)
         #[arg(long)]
         port: Option<u16>,
-        /// SMTP username
+        /// SMTP username (omit for an unauthenticated relay)
         #[arg(long)]
         username: Option<String>,
         /// From email address
@@ -25,6 +25,9 @@ pub enum ConfigCommands {
         /// From display name
         #[arg(long)]
         from_name: Option<String>,
+        /// Transport security: starttls (default), tls, or none
+        #[arg(long)]
+        tls_mode: Option<String>,
     },
     /// Show current configuration
     Show,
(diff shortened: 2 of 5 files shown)


olivierlambert:
Self-review pass. Four things I got wrong or left thin in the first commit, fixed in 9b1e93e.

**The fix had no automated test.** I verified the send path by hand and only unit-tested the `uses_auth()` predicate, which means "skip auth when the username is empty" could have regressed into "never authenticate" without a single test failing. Now that `tls_mode = none` exists, a relay is reachable in-process, so two `#[tokio::test]`s drive the real `send_email()` against a fake MTA advertising neither AUTH nor STARTTLS. Both were mutation-checked: reverting the conditional makes the first fail with the exact error from #190.

```
send should succeed against a no-auth relay:
  internal client error: No compatible authentication mechanism was found
```

**Relaxing the env block let it shadow a working database config.** This is the one that bothered me. Setting only `CALRS_SMTP_HOST` and `CALRS_SMTP_FROM_EMAIL` used to be a *partial* block that fell back to the database. After this PR it governs, so anyone in that state (the old code warned at them, it did not stop them) silently drops from authenticated to unauthenticated sends, the relay rejects the mail, and the admin form locks itself because the environment now governs. That is the same "nothing looks broken from the outside" failure the issue is about, which would have been a poor way to fix it. It now warns once per process, and only when an enabled database row actually carries a username. Still a behaviour change worth a CHANGELOG note at release.

**The TLS mode `<select>` could have marked two options `selected`.** A browser honours the last one, so an operator on `starttls` would have been shown "None, unencrypted" and downgraded their transport by saving the form untouched. The guard was right, but nothing pinned it; a test now covers all three modes, also mutation-checked.

**`none` was unreachable from a script**, which is most of the point for a containerised local relay. `calrs config smtp` gains `--tls-mode`, so the reporter's own repro line now runs with no prompts at all:

```console
$ calrs config smtp --host 127.0.0.1 --port 2526 --tls-mode none \
    --username '' --from-email noreply@example.com --from-name calrs
! TLS mode 'none' sends mail unencrypted. Only use it for a relay on this machine.
✓ SMTP configured (127.0.0.1:2526, none)

$ calrs config smtp-test guest@example.com
✓ Test email sent!

# and on the MTA side:
SERVER: message accepted (DATA completed)
SERVER: commands seen: EHLO | MAIL FROM:<noreply@example.com> | RCPT TO:<guest@example.com> | DATA | QUIT
```

Also checked and found fine: every web send path takes `load_smtp_config` through `if let Ok(Some(..))` or `.ok().flatten()`, so the new `bail!` degrades to "no SMTP" and cannot fail a booking; the two CLI sites `?`-propagate but only after the booking is committed; and the `bail!` message surfaces verbatim in the admin panel through `smtp_error` rather than just locking the form.

894 tests green, clippy clean.

PR #196, opened 2026-08-29
i18n: localize the dashboard (host-facing pages)

### LINKED ISSUE #195: i18n: localize the dashboard (host-facing pages)
## Problem

Since #59/#60/#62/#63/#76/#176, the guest side is localized end to end (booking flow, server-side dates, guest emails) in seven languages. The host side is still English-only: `dashboard_*.html`, settings, the event type / team / source forms, profile, overrides, troubleshoot.

So a host who doesn't speak English can offer guests a translated booking page, but has to manage it through an interface they can't read.

#59 deferred this on purpose ("Dashboard pages. Host-facing surface, larger scope, and gated by date-localization decisions") and shipped the language dropdown in Profile & Settings as "foundation for the dashboard translation pass". Date localization landed in #60 the same day, so the gate is cleared; the dropdown still affects zero rendered pages.

## Proposed minimal solution

Nothing new architecturally, just extend the existing Fluent setup to the host templates:

- Extract strings from the ~20 host-facing templates into namespaced keys (`dashboard-*`, `settings-*`, ...), same pattern as the booking flow.
- Locale resolution already exists: user preference, then `Accept-Language`, then English.
- Seed the other locales on Weblate, branching off `i18n` per the documented workflow.

## Acceptance criteria

- [ ] Dashboard, settings, forms, profile, overrides and troubleshoot render through Fluent keys
- [ ] The Profile & Settings language preference applies to all host-facing pages
- [ ] EN and FR at 100% (I can contribute the French strings, human-translated)
- [ ] No visible change for English users

## Out of scope

- Host-side emails (`src/email.rs`): different render path, separate pass, same as #62 did for guest emails
- CLI output, logs, API error strings
- New languages beyond the registered locales

## References

#59 (deferral note and language preference), #60 (server-side dates), #62, #63, #76, #176

olivierlambert:
Closes #195.

The host side now renders through Fluent end to end: dashboard, settings, every form, profile, overrides, troubleshoot, the auth pages, and the admin panel. English and French are both complete.

## Base branch

Opened against `i18n` per the documented workflow, but branched off `main` rather than `i18n`. `origin/i18n` is 25 commits behind and its one unique commit (`14d1bd9`, the event-type copy-link keys) is already present in `main`'s `i18n/en/main.ftl` — the two `i18n/` trees are byte-identical. Branching off the stale `i18n` would have meant localizing an older `dashboard_event_types.html` and dropping the newer strings `main` added. Say the word if you'd rather I rebase.

## How `lang` reaches a host page

The `AuthUser`, `AdminUser` and `OptionalAuthUser` extractors resolve it once, where both the user row and the request headers are already in hand: saved preference, then `Accept-Language`, then English. A dashboard render site passes `lang => auth_user.lang` and nothing else, which is what kept 34 render sites to a one-line change each. Login and register have no user row yet, so they call `detect_from_headers` directly, and their validation errors now travel as Fluent keys instead of English literals.

`AdminUser` changed from a tuple struct to `{ user, lang }`; the 30 `admin.0` uses became `admin.user`.

## Two things that would have shipped broken

**Plurals were silently wrong.** The `t()` helper stringified every argument, and a string `"1"` never matches a Fluent `[one]` variant — it falls through to `other`. The first plural key I added rendered "1 members". `t()` now hands integers to Fluent as numbers. Grouping is switched off, so an integer still renders exactly as before: "1440", never "1,440", which would have been a visible English regression on the booking-notice copy. Both behaviours have regression tests.

**Literal braces.** Fluent reads a bare `{` as the start of a placeable, so the meeting-pattern help (which documents `{username}`, `{random}`) needed `{"{"}`. Verified by rendering, not by reading the spec.

## Runtime-composed strings

Some hints are assembled in JS as the visitor edits a field, so Fluent arguments cannot serve — the values do not exist at render time. Those keys carry `%1`/`%2` placeholders substituted client-side, collected in one object per page (`ETF_I18N`, `ADMIN_I18N`) built with `{{ t('key') | tojson }}`, which is the pattern `slots.html` already used. Everything else uses real Fluent arguments.

## Guards

Three tests in `src/i18n.rs`, all of which I mutation-checked:

- every `t()` key referenced from a template exists in the English bundle
- every template still loads
- French covers every English key — this is what stops a French dashboard sprouting an English sentence the next time someone adds a key

I also rendered all 19 dashboard templates in French and confirmed no raw message id reaches the output.

## Scope notes

- **The admin panel is included** even though the issue's acceptance criteria did not list it. It was the last English-only host surface and a third of the strings; splitting it out would have left the pass visibly half-done. Easy to drop into a follow-up if you'd rather.
- **39 pre-existing French gaps are backfilled** (guest-side `book-phone-*`, `sms-*`, `booking-blocked-*`, and the event-type-form notice keys). They predate this branch, but "FR at 100%" is in the acceptance criteria and the new parity test would fail without them. FR now matches EN at 954 keys.
- **Host-side emails, CLI output and logs are untouched**, as the issue specifies.
- Two `data-confirm` guard tests needed updated expectations: the confirm sentence now comes back from Fluent and is escaped in one piece, so its own apostrophes are escaped too. The `onclick` stays static and the payload stays confined to `data-confirm` — strictly more escaping, identical after the browser decodes it.

## Unrelated flake spotted

`commands::config::tests::config_general_set_and_clear` fails intermittently: `src/caldav/mod.rs:1654` sets `CALRS_ALLOW_PRIVATE_HOSTS` process-wide while that test reads it concurrently. Present on `main`, unrelated to this branch, blocked one commit here. Worth its own issue.

## For @typovrak

The French is mine and would benefit from your review — you offered human-translated strings on the issue, and a native pass over ~500 new keys is worth more than my draft. The terminology choices worth arguing about: *Types d'événement*, *Exceptions* for overrides, *Diagnostic* for troubleshoot, *Prendre en charge* for claiming a booking, *Marge* for buffer.

## Summary by CodeRabbit

* **New Features**
  * Added Brazilian Portuguese and expanded localization across dashboard, authentication, administration, and booking interfaces.
  * Added unauthenticated SMTP relays with configurable TLS modes.
  * Added booking horizons, SMS notifications, guest phone validation, calendar-file downloads, and copy-link actions for eligible event types.

* **Bug Fixes**
  * Improved OAuth calendar handling, availability scheduling, calendar navigation, ICS generation, and environment-controlled settings.
  * Improved translated validation messages and data escaping across the interface.

* **Documentation**
  * Added Authentik OIDC SSO setup guidance and updated SMTP and security configuration references.


olivierlambert:
Review pass on this branch. Four things needed fixing; three were real bugs I introduced, one predated the branch.

## 1. Stored XSS (introduced here)

To keep word order for translators, sentences that mix text and markup render through `t(..., x='<strong>' ~ value ~ '</strong>') | safe`. `| safe` turns autoescaping off for the *whole* string, so every value spliced in has to be escaped explicitly — and I had not been.

A guest picks their own name. Booking as `<script>alert(1)</script>` put this on the host's dashboard:

```
<strong>Intro</strong> with <script>alert(1)</script>
```

Before this branch those rows were plain `{{ b.guest_name }}` and safe, so this was a regression, not a pre-existing hole. Fixed at all 21 interpolation sites; the value is now `| escape`d before concatenation, which keeps the template's own markup while escaping the data. Guarded by a test that feeds a script tag through each affected template and fails if it survives — mutation-checked by reverting one escape.

## 2. Three pages ignored the language entirely (introduced here)

Booting the server and reading real pages caught what the tests did not: **profile settings, the team form, and the write-back setup screen** rendered English no matter what the viewer had saved. Their handlers don't build context through `sidebar_context()`, which is the shape my earlier bulk pass keyed on, so they were skipped. Two render through helpers (`settings_render`, `render_team_form_error`) that now take the lang their callers already hold. The settings page's own flash and validation messages were English literals in Rust too, so those went through Fluent as well.

This fails silently — the page looks fine, just not to its reader — so there's now a test that scans the handlers and fails if a host template is rendered without a lang. It reproduces exactly what shipped.

## 3. The user-delete confirm dialog lost its paragraph breaks (introduced here)

Fluent has no backslash escapes in text, so the `\n\n` I moved into the FTL reached the browser as literal characters. Newlines now come from unicode-escape placeables; verified the rendered attribute carries real newlines.

## 4. Host-side form errors were still English (gap in the original pass)

The forms rendered in French but their validation errors did not — team, event-type and calendar-source handlers built English literals in Rust and passed them into the translated form. 21 call sites now go through Fluent. The CalDAV server's own error text stays verbatim inside the connection-failed sentence: that message is not ours to translate. Library-level error enums stay English, as the issue specifies.

Added a matching guard for Rust-side keys (a typo in `translate(lang, "key")` renders the raw id just as silently as it would from a template), also mutation-checked.

## Verification beyond the test suite

- **Ran the real server** and drove it with curl: registered, saved a French preference, and confirmed the preference beats `Accept-Language` on every dashboard page, that the success flash and validation errors come back in French, and that clearing the preference restores English.
- **Syntax-checked every rendered `<script>` block** in both languages with `node --check`: 0 errors. This matters because French is full of apostrophes; `| tojson` emits them as `'` rather than breaking out of a JS string.
- **Diffed the English render against `origin/main`** template by template. The only differences are HTML entity encoding (`&` → `&amp;`, `&mdash;` → `—`, `/` → `&#x2f;`), whitespace inside HTML, and the new JS constant lines — all of which a browser renders identically. The acceptance criterion "no visible change for English users" holds.

908 tests, clippy clean. Nothing merged.

PR #197, opened 2026-08-30
i18n: complete the six remaining locales (de, es, it, pl, pt, et)

olivierlambert:
Fills the six locales left behind by #196. **All eight now sit at 987 keys**, so the per-key English fallback exists but nothing uses it.

| locale | before | added |
|---|---:|---:|
| de | 166 | 821 |
| es | 166 | 821 |
| it | 166 | 821 |
| pl | 166 | 821 |
| pt | 184 | 803 |
| et | 0 | 987 |

~5,000 strings. Everything below is what I did to make that volume safe rather than merely large.

## Register: the trap I nearly walked into

I wrote the whole German set in formal **Sie**, then noticed the existing German says *"Deine Buchung wurde verschoben"* — informal **du**. A dashboard addressing the operator formally while the booking page addresses the guest informally is the kind of inconsistency a native speaker notices immediately. I rewrote the 122 affected strings by hand (a script cannot do it: German verb forms change with register).

That made me check the rest before writing a word. Spanish, Italian, Polish and Portuguese all use informal too (tú / tu / ty / você), so all six are informal. Estonian had no precedent, so it follows the family with *sa*. The German captcha strings are the one leftover in formal register; I left them rather than rewrite translations that came from someone else mid-pass.

## Polish plurals are not English plurals

Polish selects **one / few / many** for integers, not one/other. Copying English's two variants gives correct output at 1 and wrong output at 2 and 5. All ten plural messages carry the three categories, verified against the running bundle:

```
1 → 1 członek     2 → 2 członkowie    5 → 5 członków
22 → 22 członkowie   25 → 25 członków
```

A new test enforces this per locale, so the next person to add a plural key cannot quietly ship one/other for Polish.

## How ~5,000 strings were checked

A validator compares every locale against English on: coverage, the **set** of `{ $placeables }`, `%1`/`%2` markers, HTML tag sequence, brace/newline escapes, and plural categories. All eight report **0 errors**. It also lists values identical to English so I could eyeball them; the survivors are genuinely identical words (*Avatar*, *Gateway*, *Port*, *(optional)*).

Beyond that:

- **Rendered every dashboard page in all eight locales and ran `node --check` on every `<script>` block**: 0 syntax errors. This is the check that matters most here, because Italian, Estonian and French are full of apostrophes and `| tojson` has to emit them as `'` rather than break out of a JS string. Confirmed it does.
- **Booted the real server** and read the login page, dashboard and admin panel in de/pl/et/it. All render.
- Two guard tests, generalized from the French-only one and both mutation-checked.

910 tests, clippy clean.

## Where to point a reviewer first

**Estonian.** It was an empty file, so all 987 keys are mine and there was no existing translation to take register or terminology from. It is the locale I have least confidence in, and the one where a native pass is worth the most. The others at least had 166 keys of prior art to match.

After that, terminology is the thing worth arguing about, since I had to coin it consistently across ~500 host-side keys per language:

| | event type | override | buffer | claim | slug |
|---|---|---|---|---|---|
| de | Terminart | Ausnahme | Puffer | übernehmen | Kürzel |
| es | tipo de evento | excepción | margen | asumir | identificador |
| it | tipo di evento | eccezione | margine | prendere in carico | identificativo |
| pl | typ wydarzenia | wyjątek | bufor | przejąć | identyfikator |
| pt | tipo de evento | exceção | intervalo | assumir | identificador |
| et | sündmuse tüüp | erand | puhver | üle võtma | identifikaator |

## One consequence worth knowing

Adding a translatable key now costs eight values, not one: the coverage test fails until every locale has one. That is deliberate, and CLAUDE.md now says so along with the escape hatch (land the key on `i18n`, let Weblate fill the rest before the merge to `main`).

## Summary by CodeRabbit

- **New Features**
  - Added comprehensive translations across eight supported locales for booking flows, dashboards, settings, authentication, teams, calendars, notifications, administration, and validation messages.
  - Added localized phone fields, SMS notifications, calendar actions, booking policies, event-type settings, and troubleshooting screens.
- **Bug Fixes**
  - Improved localization consistency and message coverage across supported languages.
  - Enhanced pluralized messages and localized fallback behavior.
- **Documentation**
  - Updated localization guidance to reflect complete language coverage and translation requirements.


olivierlambert:
Ran a second pass over the branch as a reviewer rather than a translator: checking each locale against the *voice* of the English rather than against the dictionary. Four things were out of step; all fixed in `5f07852`.

First I characterised the English empirically instead of from memory. It is terse (33 characters average), imperative-heavy (144 strings open with a bare verb), and softens exactly one class of message: it says "Please" in eight validation errors and nowhere else.

**Ellipsis.** English writes `...` in sixteen strings and `…` in two. Strings translated before this branch used `...`; the ones I added used `…`. So the same locale rendered both characters on the same screen — "Loading..." next to "Generating…". Now every locale matches English per key.

**Softening.** The eight "Please …" errors had scattered treatment: fr kept it in 4, pt in 1, et in 2, and es/it/pl in none. All seven now carry it in their own idiom, so a Spanish or Polish validation error reads as gently as the English it came from.

**Duplicate strings.** Twenty-six English strings appear under more than one key, and **fourteen renderings disagreed** across six locales. "Pick an event type to book a time" is the same sentence on the user profile and the team profile, yet the two pages said different things in fr, de, es, it and pl. Same for "Create account" as heading vs button. All unified.

**German register.** My earlier informal pass missed six strings — the five captcha ones and a profile line reading *"Wählen Sie einen Veranstaltungstyp aus"*, which was both formal and used a term the other 48 occurrences don't. German is now consistently *du* and consistently *Terminart*.

### Deliberately left alone

The **em dash**. Spanish, Italian and Portuguese keep it where it separates (page titles, email subjects, the signature, provider-name prefixes) and use a colon where English uses it to introduce an explanation — which is what those languages actually do. The same sixteen keys are treated the same way in all three, so it is a rule, not drift.

The **four captcha/aria strings** where English says "please wait" / "please verify". Those are vendor-widget microcopy written by earlier translators; I did not rewrite them for a stylistic nicety. Residual inconsistency worth a translator's eye, not mine.

### Checks re-run after the changes

- Validator: **0 errors** on all seven (coverage, placeable names, `%n` markers, HTML tags, escapes, plural categories).
- Length ratios 1.07–1.24 — no bloat.
- Quotation marks correct per locale: `„…“` de/et, `„…”` pl, `«…»` fr/es/it/pt. The only straight quotes left are the Fluent `{"
"}` escapes.
- Punctuation parity vs English: the only differences are word-order endings and German `Std.` / Polish `godz.` abbreviation periods, both correct.
- Rendered every page in all eight locales; `node --check` on every script block passes, and the translated JSON `i18n` block in `slots.html` parses in all eight.

910 tests, clippy clean.

olivierlambert:
Acted on all four CodeRabbit findings in `0f6a0f7`. Three were real; one I fixed differently than proposed.

**Estonian footer** — `Tööl hoiab calrs` reverses the relationship: it says calrs is *being kept at work*, not doing the work. Now `Töötab platvormil calrs`. Good catch, that one was plainly wrong.

**Italian `settings-dynamic-group-help`** — the adjective agreement was real (`URL` is masculine in Italian tech usage). Rather than swap `estemporanee` for `estemporanei`, I restructured to `creare al volo URL di incontro collettivo`: with the adjective trailing two candidate nouns, even the corrected form leaves the reader deciding whether it modifies `URL` or `incontro collettivo`.

**Italian `book-phone-help`** — here the proposed fix changes the meaning. `se non inizi` is second person, matching the English "unless **you** start with +", and Spanish, Polish and Estonian all render it the same way. `iniziano` would move the subject to the numbers. That said the ambiguity is genuine, because `inizi` is also a valid subjunctive, so I removed the construction entirely: `senza il + iniziale si presume { $country }`, which is what the German and Portuguese already do.

**CLAUDE.md** — correct, and I had left it self-contradictory. Step 2 still said stub locales may omit keys and step 4 said only French is enforced; both stopped being true when the coverage test grew to all eight locales. Fixed, and step 4 now points at `cargo test i18n::`.

Validator still reports 0 errors on all seven; 910 tests green.

PR #198, opened 2026-08-30
Localize the guest booking flow's error responses, and fix the flaky env-var test

olivierlambert:
Two fixes I flagged during #196/#197 and left unfiled.

## 1. The guest booking flow answered in English

101 failure cases returned a bare English sentence with no page around it — "Event type not found.", "This slot is no longer available.", "This invite link has expired.", "Too many booking attempts." They predate all the localization work, and they are the reason the guest side was not really localized end to end despite #195 asserting it was.

Nineteen distinct messages: sixteen new keys plus two that already existed, translated into all eight locales. **1003 keys per locale now.**

The interesting part was where `lang` comes from. 99 of the 101 sites already had one in scope. Five sat *above* the line where their handler resolves the host's saved language — and at that point the host lookup is precisely what failed, so there is no host language to honour. Those fall back to the visitor's own `Accept-Language`, which is the only thing actually known there. `redirect_team_link_to_team` had no header at all and now takes one.

**Four strings stay English on purpose:** the CSRF rejection, the 500 page, and the OIDC failure. They live in `verify_csrf_token`, `internal_error_*` and `oidc_error_response`, called from ~240 sites between them with no `lang` to thread, and they are diagnostics rather than flow messages — the category #195 put out of scope.

Verified against the running server:

```
/u/nope        en User not found.      de Benutzer nicht gefunden.   pl Nie znaleziono użytkownika.
               et Kasutajat ei leitud. it Utente non trovato.
/team/nope     fr Équipe introuvable.  pt Equipe não encontrada.
```

## 2. The flaky test

`config_general_set_and_clear` has failed on me four times — twice blocking a local commit, once red on #197's Coverage job, once during the pre-push check on `main`.

`commands::config` already had a lock for the runtime-settings environment, and a guard to restore it. But only one of its two tests took the lock, and `caldav`'s allowlist test sets the same process-global variable from a module that cannot see it. A module-local lock cannot serialise against another module.

Both now live in `crate::test_support` and all three tests share them. The sync caldav test uses `blocking_lock()`, which is safe there because a plain `#[test]` has no runtime.

Measured rather than assumed — running the two tests together, 12 times each way:

| | failures in 12 runs |
|---|---:|
| without the lock | **5** |
| with the lock | **0** |

## Notes

- CLAUDE.md now records the rule: any test touching `CALRS_*` takes `ENV_LOCK` and holds an `EnvGuard`, because a module-local lock is what caused this.
- 910 tests, clippy clean.

## Summary by CodeRabbit

- **New Features**
  - Added localized guest booking error messages for unavailable slots, invalid dates and times, invite-link issues, rate limits, unavailable team members, restricted event types, and missing users.
  - Booking and rescheduling error responses now use the visitor’s language when available.
  - Added and updated German, English, Spanish, Estonian, French, Italian, Polish, and Portuguese translations, including phone, SMS, cancellation, and event settings text.
- **Documentation**
  - Updated localization coverage notes and documented intentional English-only messages.


olivierlambert:
Both CodeRabbit findings were valid; fixed in `092382c`.

**The variable-held message was a real miss on my part.** My pass matched `Html("literal".to_string())`, so it walked straight past a site that picks its message with an `if` and passes the binding. A private team's event page was still answering in English. Now it picks a Fluent key instead of a sentence, and the access-control branch is untouched — a public event on a private team still answers "not found" rather than admitting the event exists. I re-scanned for the pattern: that was the only one, and there are now zero `Html(<var>.to_string())` sites.

**The French agreement is also real.** `{ $country } est supposé` cannot agree with an interpolated country name — *la France est supposée*, *le Portugal est supposé*. Rather than pick a gender, I restructured to avoid agreeing with the placeholder at all (`sans le + initial, nous supposons { $country }`), which is what the other six locales already do. Worth noting this line predates the branch; CodeRabbit caught it because the file changed.

### One thing for you to decide

While re-scanning I found an adjacent category I deliberately did **not** touch, because you scoped this to the bare responses:

`render_booking_action_error` already resolves `lang` and renders a properly localized page — but its 22 call sites pass English `title` and `message` literals. So a guest hitting an expired approve/decline/cancel link, or tripping booking validation ("Notes must be 5000 characters or less.", "Additional guests are not allowed for this event type.", "Cannot book more than one year in advance.") gets a nicely-styled page with English text inside it.

Roughly 25–30 distinct strings, one clean mechanism, so about 240 translations. Arguably more visible than the bare responses I just fixed, since these are the paths with page chrome. Say the word and I'll do it in a follow-up.

Issue #200, opened 2026-08-30
Weblate 404


typovrak:
Hello, i saw this link multiple times in your repository but on my side, i have a 404, is this normal ?

https://hosted.weblate.org/projects/calrs/

<img width="1920" height="1020" alt="Image" src="https://github.com/user-attachments/assets/bd4f2c72-f44e-4aa1-bfcc-ca77c4486b41" />

olivierlambert:
Despite asking to be hosted as an open source project, Weblate never answered and closed the project. I will see what to do next.

olivierlambert:
Thanks for reporting this, and sorry you hit a 404 chasing a link we were advertising. It was not normal, and it was worse than a stale link.

Digging into it: the Hosted Weblate project was applied for and never approved, so the account was closed. All three URLs in the README were dead, including the status badge, which meant the Localization section was rendering a broken image. Checking the history, **no commit in this repository was ever authored by Weblate** and there is no `.weblate` config, so the integration never worked at a single point in time. A CHANGELOG line crediting "Translation updates from Weblate" turned out to be a maintainer's own branch merge.

Meanwhile six people have contributed translations by ordinary pull request. So rather than chase a replacement platform, the flow that has actually been delivering translations all along is now the documented one.

Fixed in #203:

- Every Weblate link removed from the README and from the website
- A real contributor guide in its place: how to improve a language that already ships, how to add a new one, and what an LLM reliably gets wrong when seeding a locale (Fluent placeables and brace escaping, plural categories that are not English's two, and the formal register models default to in German and Spanish)
- Two stale instructions fixed that would have walled anyone following them. "Create the file empty, the runtime falls back to English" stopped being true once the coverage test landed, so registering an empty locale now fails with 1043 errors. And a step that was documented nowhere: a new locale has to declare its CLDR plural categories or the build fails with no hint why.

The guide is at https://github.com/olivierlambert/calrs#contributing-a-translation. If you were thinking of translating something, that page should now be accurate. Shout if any step does not match what you actually hit.

PR #205, opened 2026-08-30
fix(settings): confirm a language change in the language just saved

### LINKED ISSUE #204: i18n: the settings page answers a language change in the old language
## Problem

Switching the dashboard language applies everywhere except on the response to the switch itself:

1. Log in with the browser in English, open Dashboard → Settings.
2. Set Language to "Français", save.
3. The page that comes back is still in English, including the "Settings saved" notice. Reload: French.

Tested on v1.17.1; the code on `main` matches.

## Where it comes from

`settings_save` (src/web/mod.rs) re-fetches the user row after the UPDATE "to show updated values", but renders with `auth_user.lang`, which the extractor resolved before the save. The handler translates the success and error flashes with the same stale value.

No other handler is affected: every other page resolves the language on its own request, after the preference is already saved.

## Suggested fix

Resolve the language a second time after the re-fetch, from the fresh row and the request headers already in the handler's scope, and hand that to `settings_render` and the flash translation. A post/redirect/get would fix it too, but it drags the flash message along and that is more change than the bug needs.

typovrak:
Closes #204.

## The bug

Saving "Français" from an English session returns an English page, "Settings saved" included. Only a reload shows the French one. The preference itself is stored correctly, so every later page is right; it is the confirmation that lags.

## Cause

`settings_save` re-fetches the user row after the UPDATE to show the new values, but renders with `auth_user.lang`, which the extractor resolved from the pre-save row. The success and error flashes translate against the same stale value. The one request where the language changes is the one request that cannot see it.

## Fix

Resolve the language a second time after the re-fetch, from the fresh row and the request headers already in the handler's scope:

```rust
let lang = crate::i18n::resolve(updated_user.language.as_deref(), &headers);
```

I considered a post/redirect/get instead, which would fix it structurally, but it drags the flash message along and that is more change than this bug needs.

Nothing else is affected: `settings_save` is the only handler that writes `users.language`, and every other page resolves on its own request, after the preference is stored.

## Test

`settings_save_confirms_language_change_in_the_new_language` drives a real POST with `language=fr` under an English session and asserts the French confirmation is present and the English one is not.

Mutation-checked, per the standard the 1.17.1 notes set: restoring `auth_user.lang` on the render fails it with "confirmation should render in the language that was just saved". The negative assertion matters on its own, since a page carrying both strings would otherwise pass.

## Checks

- `cargo test`: 914 passed, 0 failed
- `cargo clippy --all-targets`: clean
- `cargo fmt --check`: clean
- Manually on a local build, all eight locales available: en to fr, fr back to en, and "Auto" falling through to `Accept-Language`

No new Fluent keys, so this targets `main` rather than `i18n`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Settings changes are now confirmed in the newly selected language.
  * Added coverage to verify that French language selections display confirmation messages in French.

### DIFF
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -3443,16 +3443,15 @@ async fn settings_save(
             let updated_user = crate::auth::get_user_by_id(&state.pool, &user.id)
                 .await
                 .unwrap_or_else(|| user.clone());
+            // auth_user.lang was resolved before the UPDATE, so a language
+            // change would confirm in the old language without this.
+            let lang = crate::i18n::resolve(updated_user.language.as_deref(), &headers);
             let sidebar = sidebar_context(&auth_user, "settings");
             settings_render(
                 &state,
                 &updated_user,
-                auth_user.lang,
-                Some(&crate::i18n::translate(
-                    auth_user.lang,
-                    "settings-saved",
-                    None,
-                )),
+                lang,
+                Some(&crate::i18n::translate(lang, "settings-saved", None)),
                 None,
                 sidebar,
                 imp,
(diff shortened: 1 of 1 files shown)


olivierlambert:
Merged, thank you. Clean diagnosis and the smallest fix that closes it.

I checked the parts a diff does not show: `sidebar_context` carries no translated strings, so the whole page follows the new language and not just the flash; `settings_save` is the only writer of `users.language`, so no other handler has this staleness; and leaving `auth_user.lang` on the error paths is right, since nothing was persisted there. Locally: 914 tests, clippy and fmt clean, and the mutation check fails exactly as you described.

Note for your next PR: CI did not run here until I approved it by hand. First contributions from a fork sit in `action_required`, so a green CodeRabbit is not a green build. Nothing for you to do, just so the empty checks list does not worry you.

One thing your patch made visible: the validation error paths re-render from the stored row rather than the submitted form, so a bad name silently discards the rest of the edits. Filed as #206, out of scope here.

If you run calrs day to day, field reports on the language switching across the eight locales are welcome.

olivierlambert:
This shipped in [calrs 1.18.0](https://github.com/olivierlambert/calrs/releases/tag/v1.18.0). Thanks @typovrak for the fix, and for the Weblate report (#200) too!

If you get a chance, could you upgrade to 1.18.0 and confirm the settings page behaves as expected on your side, with nothing else regressed? This release includes migration 064, so please back up your database first.

PR #207, opened 2026-08-31
fix(settings): keep the submitted values when a field is rejected

### LINKED ISSUE #206: settings: a validation error discards the rest of the form
## Problem

Open Dashboard → Settings, edit several fields at once (title, bio, timezone, language), then clear the name field and save. The page comes back with the name error, and every other field has reverted to its stored value. The edits are gone. The name field itself shows the old, valid name while the error says the name is invalid.

Nothing is lost in the database, the UPDATE never ran. It is the typing that is lost, silently.

## Where it comes from

The three non-success paths in `settings_save` re-render with `user`, the row the extractor read before the request, rather than with the submitted form:

- name length, `src/web/mod.rs:3264`
- booking email, `src/web/mod.rs:3382`
- UPDATE failure, `src/web/mod.rs:3472`

`settings_render` derives `form_name`, `form_title`, `form_bio`, `form_booking_email`, `form_timezone`, `form_language` and `allow_dynamic_group` from that row. Two fields are the exception and do come from the submitted form, `avail_schedule` and `lend_resource_write`, so the re-rendered page is half submitted values and half stored ones.

## Suggested fix

Have the error paths render what was submitted, so the user only has to correct the field the error names, and so the offending value is the one shown. Either pass the form fields to `settings_render` in place of the row, or give it a small struct that the success path fills from the fresh row and the error paths fill from the form.

## Notes

Not a regression. Found while reviewing #205, which fixed the neighbouring staleness bug on the success path (#204).

olivierlambert:
Closes #206.

## The bug

Edit several fields on Profile & Settings, clear the name, and save. The name error comes back and every other edit is gone: title, bio, booking email, timezone, language and the dynamic-group checkbox have all reverted to the stored row. The name field shows the old, valid name while the error says the name is invalid.

The username is the sharp edge. It is saved before the later checks run, so a bad booking email showed the previous username back after the new one had already been written to the database.

## Cause

All five non-success paths in `settings_save` re-rendered from `user`, the row read before the request. `settings_render` derived every `form_*` value from it. Only `avail_schedule` and `lend_resource_write` came from the submitted form, which is why the page was half one and half the other.

Each of those paths repeated the same eleven-line call, which is how they drifted apart, and is the same shape of duplication behind #204.

## Fix

`SettingsValues` holds the editable half of the form. The GET and the success path build it from the row via `from_user`; the error paths build it from what was submitted. Every field is now normalised before the first check, so a rejected one can be sent back alongside the others, and the five repeated call sites collapse into one `render_error` closure.

Two deliberate choices:

- The page stays in the pre-request language on an error. Nothing was saved, so the stored preference is still the user's language. Only the success path re-resolves, as #205 made it do.
- An empty username field means "keep the current one" to the handler, so that is what the form shows back rather than an empty box.

Also drops a redundant `sidebar_context` rebuild in the success arm.

## Tests

Both mutation-checked by pointing the error paths back at `from_user`:

- `settings_validation_error_keeps_the_submitted_fields` drives an empty name with five other fields edited, none matching the seeded row, and asserts each comes back, including an unchecked box staying unchecked while the stored row still has it on. Fails with "a rejected field should not discard the other edits, missing value=\"Head of Demos\"".
- `settings_booking_email_error_shows_the_rejected_value` asserts the field in error shows the value that was refused. Fails with "the field in error should show the value that was rejected".

The other six settings tests pass under that mutation, which is what makes it a faithful stand-in for the old behaviour.

## Checks

- `cargo test`: 916 passed, 0 failed
- `cargo clippy --all-targets -- -D warnings`: clean
- `cargo fmt --check`: clean

No new Fluent keys, so this targets `main` rather than `i18n`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Settings forms now retain all submitted values when validation errors occur.
  * Corrected an issue where rejected fields could revert other form entries to previously saved values.
  * Avatar initials now update to match the submitted display name, including when a save is rejected.
  * Stored settings values are displayed consistently when opening or successfully saving the form.
* **Tests**
  * Added coverage to verify submitted values and corresponding avatar initials remain visible after validation failures.

### DIFF
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -3255,7 +3255,14 @@ fn settings_render(
             lang => lang,
             sidebar => sidebar,
             form_name => values.name,
-            form_initials => compute_initials(&user.name),
+            // Follows the name being shown, so an edited name updates the
+            // avatar preview too. A rejected empty name falls back to the
+            // stored one rather than rendering the "?" placeholder.
+            form_initials => compute_initials(if values.name.trim().is_empty() {
+                &user.name
+            } else {
+                values.name
+            }),
             form_title => values.title,
             form_bio => values.bio,
             form_booking_email => values.booking_email,
@@ -30063,7 +30070,8 @@ mod tests {
         let (app, _pool, session, _) = setup_test_app().await;
         let csrf = "test-csrf-settings-email";
         let body = format!(
-            "_csrf={}&name=Test+User&username=testuser&title=Head+of+Demos&booking_email=not-an-email",
+            "_csrf={}&name=Alice+Martin&username=testuser&title=Head+of+Demos\
+             &booking_email=not-an-email",
             csrf
         );
         let response = app
@@ -30084,6 +30092,21 @@ mod tests {
         assert!(
             resp_body.contains(r#"value="Head of Demos""#),
             "the other edits should survive the error"
+        );
+        assert!(
+            resp_body.contains(r#"value="Alice Martin""#),
+            "the submitted name should survive the error"
+        );
+        let avatar = resp_body
+            .split(r#"flex-shrink: 0; overflow: hidden;">"#)
+            .nth(1)
+            .expect("the avatar preview should render")
+            .split("</div>")
+            .next()
+            .unwrap_or("");
+        assert!(
+            avatar.contains("AM"),
+            "the avatar initials should follow the name being shown, got {avatar:?}"
         );
     }
 
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -2957,17 +2957,17 @@ async fn settings_page(
             .fetch_one(&state.pool)
             .await
             .unwrap_or(0);
+    let values = SettingsValues::from_user(&auth_user.user, &avail, lend_resource_write != 0);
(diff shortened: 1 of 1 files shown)


olivierlambert:
Both CodeRabbit points checked. One taken, one split out.

**Stale initials, fixed here** (ac65cc4). The avatar preview kept rendering the stored name while the input beside it showed the edited one. It now follows the name being shown, with a fallback to the stored name when the submitted one is the rejected empty string, so the circle never degrades to the "?" placeholder while the user is fixing the field. The booking-email test submits a changed name and pins both the echoed value and the initials, and it fails on the initials assertion when the fallback is reverted.

**Username written before a later failure: real, pre-existing, filed as #208.** `settings_save` writes the username in its own statement as soon as that field validates, so a booking-email error after it leaves the write standing and the page reports a save that did not happen. It predates this branch and the fix is a change to when the row is written rather than to what is rendered, so it wants its own review.

Worth noting this PR makes that bug less misleading rather than worse. The error page now shows the submitted username, which is what is actually stored, where before it showed the old one after the new one had already been written.

